PR #83, opened 2015-06-05
Attempting to fix warning from clang

mothran:
Because Arch ships with gcc 5.1.0 and causes the following error when building the pin libs:
> ../../../source/include/pin/compiler_version_check2.H:102:2: error: #error The C++ ABI of your compiler does not match the ABI of the pin kit.

I tried to build with clang and hit a few minor errors.  I am unsure about if REG_valid() will correctly error when passed a -1 value.

original error with clang:
> src/contextHandler/PINContextHandler.cpp:37:11: error: comparison of constant -1 with expression of type 'REG' is always false [-Werror,-Wtautological-constant-out-of-range-compare]


JonathanSalwan:
Thanks you're right. According to the [Pin documentation](https://software.intel.com/sites/landingpage/pintool/docs/71313/Pin/html/group__REG__CPU__IA32.html), we must use REG_INVALID_ enum instead of the constant value -1.

PR #97, opened 2015-06-19
Get only one syscalls table file.

### LINKED ISSUE #96: Build fails when multiple architectures of libc are installed
I'm trying to add triton to my ctf-tools repo (https://github.com/zardus/ctf-tools), and am getting the following error:

```ShellSession
yans@lath ~/tools/triton/pin-2.14-71313-gcc.4.4.7-linux/source/tools/Triton/build $ cmake ..
CMake Error at CMakeLists.txt:59 (if):
  if given arguments:

    "NOT" "EXISTS" "/usr/include/x86_64-linux-gnu/asm/unistd_64.h" "/usr/include/i386-linux-gnu/asm/unistd_64.h"

  Unknown arguments specified


-- Configuring incomplete, errors occurred!
See also "/home/yans/tools/triton/pin-2.14-71313-gcc.4.4.7-linux/source/tools/Triton/build/CMakeFiles/CMakeOutput.log".
```

This seems to be because I have both `linux-libc-dev:amd64` and `linux-libc-dev:i386` installed (under Ubuntu 14.04), and they each contain a `unistd_64.h` file. In turn, your recursive check for unistd_64.h returns both files, which confuses cmake.

I'll open a PR with a hackish fix, in case that's good enough.

zardus:
This should fix #96. Another solution would be to have it just search in the `/usr/include/x86_64-linux-gnu` directory. That would require the installation of 64-bit libc dev headers (whereas now, the 32-bit ones will suffice for this specific check since they also seem to include the 64-bit headers), so I went with this option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,9 +51,10 @@ link_directories("${PIN_ROOT}/extras/xed-intel64/lib")
 # This following code tries to find the unistd_64.h header.
 FILE(
     GLOB_RECURSE
-    syscalls_table_file
+    syscalls_table_files
     /usr/include/*unistd_64.h
 )
+LIST(GET syscalls_table_files 0 syscalls_table_file)
 
 # If the unistd_64.h is not found, we exit
 if (NOT EXISTS ${syscalls_table_file})


JonathanSalwan:
Thanks :)

PR #115, opened 2015-06-30
Fixed comments in examples/, tools/: ./pin -t to ./triton

Jinmo:
Since there are standalone binary "./triton", ../../../pin -t ./triton.so -script blabla -- can be removed.


JonathanSalwan:
Thanks :)

PR #120, opened 2015-07-04
Handle UNSET mem/reg

romainthomas:
We can now create symbolic variable even if there are UNSET.


src/symbolicEngine/symbolicEngine.cpp
```diff
@@ -271,18 +271,55 @@ SymbolicVariable *SymbolicEngine::convertExprToSymVar(uint64 exprId, uint64 symV
...
+{
+    SymbolicExpression    *expression = nullptr;
+    SymbolicVariable      *symVar  = nullptr;
+    std::stringstream     newExpr;
+    std::stringstream     expr("#0");
+
+    expression = this->newSymbolicExpression(expr,symVarComment);
+    this->addMemoryReference(memAddr, expression->getID());
+
+    if (expression == nullptr)
```
JonathanSalwan: This check must be applied before the line 282.

src/symbolicEngine/symbolicEngine.cpp
```diff
@@ -310,10 +347,15 @@ SymbolicVariable *SymbolicEngine::convertRegToSymVar(uint64 regId, uint64 symVar
...
+  if (regSymId == UNSET){
+    std::stringstream     expr("#0");
+    expression = this->newSymbolicExpression(expr, symVarComment);
+    this->symbolicReg[regId] = expression->getID();
+
+  }else{
+    expression = this->getExpressionFromId(regSymId);
+  }
 
   if (expression == nullptr)
```
JonathanSalwan: This check must be applied before the line 354.

src/symbolicEngine/symbolicEngine.cpp
```diff
@@ -310,10 +347,15 @@ SymbolicVariable *SymbolicEngine::convertRegToSymVar(uint64 regId, uint64 symVar
     throw std::runtime_error("SymbolicEngine::convertRegToSymVar() - Invalid register ID");
 
   regSymId = this->getRegSymbolicID(regId);
-  if (regSymId == UNSET)
-    throw std::runtime_error("SymbolicEngine::convertRegToSymVar() - This register ID is UNSET");
 
-  expression = this->getExpressionFromId(regSymId);
+  if (regSymId == UNSET){
+    std::stringstream     expr("#0");
```
JonathanSalwan: This is not the `#0` but the `symVarName` ?

src/symbolicEngine/symbolicEngine.cpp
```diff
@@ -271,18 +271,55 @@ SymbolicVariable *SymbolicEngine::convertExprToSymVar(uint64 exprId, uint64 symV
 }
 
 
+SymbolicVariable *SymbolicEngine::convertMemByteToSymVar(uint64 memAddr, std::string symVarComment)
+{
+    SymbolicExpression    *expression = nullptr;
+    SymbolicVariable      *symVar  = nullptr;
+    std::stringstream     newExpr;
+    std::stringstream     expr("#0");
```
JonathanSalwan: This is not the `#0` but the `symVarName` ?

src/symbolicEngine/symbolicEngine.cpp
romainthomas: newSymbolicExpression(std::stringstream &src) takes a source id in parameter but as we create a new symbolic variable we don't have source id so I took the #0 value.

JonathanSalwan:
Note: We will add this feature after the merge of the new symbolic engine's design.

PR #154, opened 2015-07-21
Patched/0721

cheesechoi:
save past code. It need modifying to newer fucntion.


cheesechoi:
oh sorry, i had some mistake.

PR #307, opened 2016-03-07
Added a Dockerfile to generate a container with Ubuntu 14.04 with Triton

pmarkowsky:
Added a simple Docker container to allow people to try out Triton on Ubuntu 14.04. This grabs sets up a basic enviornment with clang-3.6.2 and pulls Boost from Ubuntu's packages and then builds Z3, Capstone, and Triton from source. 

Pin is also pulled down to */opt/pin-2.14-71313-gcc.4.4.7-linux*

Triton's built in */opt/Triton-master/build* or */opt/Triton/build* depending upon if you use the release link or git.

This makes setting up a development environment with Triton as simple as 

```
[ peterm@ubuntu ( 7:24AM) ~/dockerfiles/triton  ]                                  
$  docker build --tag=triton . 
```

Where you can then run it by

```
[ peterm@ubuntu ( 7:32AM) ~/dockerfiles/triton  ]                                  
$  docker run -it --rm trition
root@62e5de69f584:/# python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import triton
>>>
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04
+FROM ubuntu:15.10
 MAINTAINER Pete Markowsky <pete@markowsky.us>
 RUN apt-get update && apt-get dist-upgrade -y && \
     apt-get install -y git cmake build-essential clang ca-certificates curl \
@@ -25,7 +25,7 @@ RUN cd /opt && curl -o pin.tgz -L http://software.intel.com/sites/landingpage/pi
 # RUN cd /opt && git clone https://github.com/JonathanSalwan/Triton.git && \
 #     cd Triton && mkdir build && cd build && cmake -G "Unix Makefiles" .. && \
 #     make install &&  cd .. && python setup.py install
-RUN cd /opt && \
+RUN cd /opt/pin-2.14-71313-gcc.4.4.7-linux/source/tools/ && \
    curl -o master.zip -L https://github.com/JonathanSalwan/Triton/archive/master.zip && unzip master.zip && cd Triton-master/ && mkdir build && cd build && \
-   cmake -G "Unix Makefiles" .. && make install && cd .. && python setup.py install
+   cmake -G "Unix Makefiles" -DPINTOOL=yes -DKERNEL4=yes .. && make install && cd .. && python setup.py install
 ENTRYPOINT /bin/bash
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM ubuntu:14.04
+MAINTAINER Pete Markowsky <pete@markowsky.us>
+RUN apt-get update && apt-get dist-upgrade -y && \
+    apt-get install -y git cmake build-essential clang ca-certificates curl \
+    unzip libboost-dev python-dev python-pip && apt-get clean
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM ubuntu:14.04
```
JonathanSalwan: Is ubuntu 14.04 using a kernel 4.x ?

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# Install pintool
+RUN cd /opt && curl -o pin.tgz -L http://software.intel.com/sites/landingpage/pintool/downloads/pin-2.14-71313-gcc.4.4.7-linux.tar.gz && tar zxf pin.tgz
+
+# now install Triton
+# uncomment below to pull form git
+# RUN cd /opt && git clone https://github.com/JonathanSalwan/Triton.git && \
+#     cd Triton && mkdir build && cd build && cmake -G "Unix Makefiles" .. && \
+#     make install &&  cd .. && python setup.py install
+RUN cd /opt && \
```
JonathanSalwan: As you downloaded Pin into `/opt`, you should install Triton into `/opt/pin-2.14-71313-gcc.4.4.7-linux/source/tools/`

Dockerfile
pmarkowsky: It's not. However Docker uses the host OS' kernel. So depending upon what
you host this container in you'll get that version of  the kernel.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+   curl -o master.zip -L https://github.com/JonathanSalwan/Triton/archive/master.zip && unzip master.zip && cd Triton-master/ && mkdir build && cd build && \
+   cmake -G "Unix Makefiles" .. && make install && cd .. && python setup.py install
```
JonathanSalwan: If it's a kernel 4.x, please use `cmake -DPINTOOL=yes -DKERNEL4=yes ..`, if it's not a kernel 4.x, you can remove the `-DKERNEL` flag but keep the `-DPINTOOL` one :)

JonathanSalwan:
Thanks for this PR Pete, it's really useful :). I think we can use `cmake -DPINTOOL=yes -DKERNEL4=yes ..` even if it's a kernel `3.x`, then we should install Triton into `/opt/pin-2.14-71313-gcc.4.4.7-linux/source/tools/`, then I can accept the PR :)

pmarkowsky:
Great, let me fix it and test. Was also realizing we could easily change
this to Ubuntu 15.10 for newer versions of clang etc.

That would let Z3 build with openmp support.

JonathanSalwan:
Great, thank you!

pmarkowsky:
As a heads up I've not had a chance to test this as thoroughly as I'd like. It builds correctly and Triton can be installed but I've not had a chance to run some scripts through it or test the pintool on anything.

In short I'd hold off on accepting this PR until tomorrow unless you'd like to test it first.

JonathanSalwan:
Looks good. Thanks!

Issue #327, opened 2016-03-31
References and SMT-LIB v2 SSA form


apach301:
For instructions, operating with memory, some symbolic references generated wrong (missing counter and determination).
Example:
> 4004f1: mov dword ptr [rbp - 0x14], edi
ref!13 = ((_ extract 31 24) (_ bv1 32)) ; byte reference - MOV operation
ref!14 = ((_ extract 23 16) (_ bv1 32)) ; byte reference - MOV operation
ref!15 = ((_ extract 15 8) (_ bv1 32)) ; byte reference - MOV operation
ref!16 = ((_ extract 7 0) (_ bv1 32)) ; byte reference - MOV operation
ref!18 = (_ bv4195572 64) ; Program Counter
> 

[trace.txt](https://github.com/JonathanSalwan/Triton/files/197272/trace.txt)
[code.txt](https://github.com/JonathanSalwan/Triton/files/197278/code.txt)

JonathanSalwan:
Are you talking about the `ref!17`? If yes, it's not a bug, this reference (17) really exist but it's not linked with the instruction. This reference is used for several internal stuffs. Basically, this reference is the concatenation of references 13, 14, 15 and 16.

apach301:
The lack of `ref!17` don't causes errors when solving expression in SMT solver, because there is no any reference linked with `ref!17`. But in some cases it happens (example.txt): `ref!40`..`ref!45` linked with `ref!39`, which don't determined.
[example.txt](https://github.com/JonathanSalwan/Triton/files/197358/example.txt)

JonathanSalwan:
Ok I see your point, as I said, `ref!17` like `ref!39` exist but are not linked with the instruction. I agree with you that if you try to reconstruct the SSA with your eyes that's not really possible because your `ref!39` is not printed but that's doesn't mean that the Triton's SSA doesn't work :)

If you call `getSymbolicExpressionFromId(17)` and `getSymbolicExpressionFromId(39)` at the end of the execution, you will see that these references really exist.

This behavior occurs only when the instruction contains a memory access and the semantics look like this:

- byte ref expression (linked with the instruction)
- [...]
- byte ref expression (linked with the instruction)
- concatenation of byte ref expression (**not linked** with the instruction, used for internal stuffs)
- flags expression if exist (linked with the instruction)
- control flow expression (linked with the instruction)

[...]

PR #328, opened 2016-03-31
#327: Add memory concatenation

enzet:
This line was deleted by 66b005c, but probably this operation still useful for trace consistency when we want to get raw trace using `getSymbolicExpressions()`.


JonathanSalwan:
Can you please rebase the master branch?

PR #341, opened 2016-06-03
Fix Windows compatibility 

illera88:
VS2013 don't support User-defined literals https://blogs.msdn.microsoft.com/vcblog/2014/08/21/c1114-features-in-visual-studio-14-ctp3/

VS2015 does support it but PIN or DynamoRio does not support VS2015


JonathanSalwan:
Hey,

Thanks, I really appreciate feedback about Windows' compatibility :).

PR #348, opened 2016-06-27
Immediate operand conversion

enzet:
Triton creates constraint

```lisp
ref!356633 = ((_ zero_extend 32) ((_ extract 31 0) (bvadd ((_ zero_extend 32) (_ bv18446744073709551568 32)) (bvadd ((_ extract 63 0) ref!356631) (bvmul (_ bv0 64) ((_ zero_extend 32) (_ bv1 32))))))) ; LEA        operation
```

with some strange constant `(_ bv18446744073709551568 32)` (18446744073709551568 = 2<sup>64</sup> - 48 > 2<sup>32</sup>).

Immediate value in Capstone has type `int64_t`, but `ImmediateOperand` takes `uint64` value. If value size is less than `QWORD_SIZE`, it may be too big. Proposed conversion creates more appropriate constant `(_ bv4294967248 32)` in

```lisp
ref!356663 = ((_ zero_extend 32) ((_ extract 31 0) (bvadd ((_ zero_extend 32) (_ bv4294967248 32)) (bvadd ((_ extract 63 0) ref!356661) (bvmul (_ bv0 64) ((_ zero_extend 32) (_ bv1 32))))))) ; LEA operation
```

I'm not sure fix is right. May be we should have a deal with signed constants instead of unsigned?

### DIFF
--- a/src/libtriton/arch/immediateOperand.cpp
+++ b/src/libtriton/arch/immediateOperand.cpp
@@ -19,13 +19,26 @@ namespace triton {
 
 
     ImmediateOperand::ImmediateOperand(triton::uint64 value, triton::uint32 size /* bytes */) {
-      this->value = value;
 
       if (size == 0)
         throw std::runtime_error("ImmediateOperand::ImmediateOperand(): size cannot be zero.");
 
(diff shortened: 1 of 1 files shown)


JonathanSalwan:
You're right!

> I'm not sure fix is right. May be we should have a deal with signed constants instead of unsigned?

Sounds good to me. ATM I prefer to deal with unsigned instead of signed.

Thanks for this fix :)

PR #374, opened 2016-08-10
Fix in assignSymbolicExpressionToMemory removing previous alignedMemo…

0ca:
The function `assignSymbolicExpressionToMemory` is not removing the previous `alignedMemoryReferences`.

So if the `ALIGNED_MEMORY` optimization is enable, and we use `assignSymbolicExpressionToMemory` to set a SE to a memory, this SE is ignored when the buildSemantics finds the previous memory reference in the vector `alignedMemoryReferences`.

This commit should fix this behavior.


JonathanSalwan:
Thanks for the patch! However, I think we should have one PR per issue. I can't merge this patch because the second patch (issue #375) is not correct.

PR #386, opened 2016-09-13
Add getter to obtain tainted memory address

illera88:


JonathanSalwan:
cf ccfcc888c309ead3cac55d0fd286d6f95ad57a7b

PR #392, opened 2016-09-26
Third party libraries not needed if building static lib

illera88:


JonathanSalwan:
You forgot the `PYTHON_LIBRARIES` btw :)

PR #402, opened 2016-10-15
Create isRegisterSymbolic and isMemorySymbolic

illera88:


src/libtriton/includes/api.hpp
```diff
@@ -459,6 +459,12 @@ namespace triton {
         //! [**symbolic api**] - Returns true if the symbolic optimization is enabled.
         bool isSymbolicOptimizationEnabled(enum triton::engines::symbolic::optimization_e opti);
 
+		//! [**symbolic api**] - Returns true if the register is symbolic
```
JonathanSalwan: In this file, indentation must be 8 spaces (no tab).

src/libtriton/api/api.cpp
```diff
@@ -929,7 +929,20 @@ namespace triton {
     return this->symbolic->isOptimizationEnabled(opti);
   }
 
+  bool API::isMemorySymbolic(const triton::arch::MemoryAccess& mem) const{
```
JonathanSalwan: Please rename `isMemorySymbolic` to `isMemorySymbolized`.

src/libtriton/api/api.cpp
```diff
@@ -929,7 +929,20 @@ namespace triton {
...
+	  this->checkSymbolic();
+	  triton::usize symbId = triton::api.getSymbolicMemoryId(mem.getAddress());
```
JonathanSalwan: You should use `this->` instead of `triton::api.` What if `mem.getSize()` is more than a `BYTE`?

src/libtriton/api/api.cpp
```diff
@@ -929,7 +929,20 @@ namespace triton {
...
+	  triton::engines::symbolic::SymbolicExpression* symbExp = triton::api.getSymbolicExpressionFromId(symbId);
```
JonathanSalwan: You should use `this->` instead of `triton::api.`. Here, there is an exception raised if `symId` is not valid - you have to manage this case`.

src/libtriton/api/api.cpp
```diff
@@ -929,7 +929,20 @@ namespace triton {
...
+	  return symbExp->isSymbolized();
+  }
 
+  bool API::isRegisterSymbolic(const triton::arch::Register& reg) const{
+	  this->checkSymbolic();
+	  triton::usize symbId = triton::api.getSymbolicRegisterId(reg);
+	  triton::engines::symbolic::SymbolicExpression* symbExp = triton::api.getSymbolicExpressionFromId(symbId);
```
JonathanSalwan: Same as the previous function. You have to manage the case where `symbId` is `UNSET` and use `this->` instead of `triton::api`.

illera88:
I did the C++ part. I let you finishing the python bindings.
Cheers

src/libtriton/api/api.cpp
```diff
@@ -929,7 +929,33 @@ namespace triton {
     return this->symbolic->isOptimizationEnabled(opti);
   }
 
+  bool API::isMemorySymbolized(const triton::arch::MemoryAccess& mem) const{
+	  this->checkSymbolic();
+	  for (int i = 0; i < mem.getSize(); i++){
```
JonathanSalwan: `int i` must be `triton::uint32`. cf [travis](https://travis-ci.org/JonathanSalwan/Triton/jobs/168097209#L1184).

illera88:
done

JonathanSalwan:
Still not :)

Issue #413, opened 2016-11-29
Semantics for 3-operand SHLD/SHRD missing


newgre:
The framework is unable to process 3-operand SHLD/SHRD instructions.

The following python snippet demonstrates the issue:

```python
instr = Instruction()
instr.setOpcodes("\x66\x44\x0F\xA4\xDD\x3C")  # shld bp, r11w, 0x3c
instr.setAddress(0x1000)
if not processing(instr):
  raise Exception("Unsupported instruction: %s" % (instr))
```

Same goes for the SHRD equivalent, e.g.,

```python
instr = Instruction()
instr.setOpcodes("\x66\x0F\xAC\xD8\x46")  # shrd ax, bx, 46h
instr.setAddress(0x1000)
if not processing(instr):
  raise Exception("Unsupported instruction: %s" % (instr))
```

JonathanSalwan:
Indeed. Let me few days to dev this. Next week I will be a little bit busy without internet.

JonathanSalwan:
`SHLD` done via 84d110bdf3a3b54d53e48d757dc19fd6193b4e5e and 5a384f9d082ede7deeb5147aca204de35ad08f59

PR #430, opened 2016-12-24
add shld semantics

gyorokpeter:
Also includes some fixes for mingw build.

### DIFF
--- a/src/examples/cpp/shld_test.cpp
+++ b/src/examples/cpp/shld_test.cpp
@@ -1,0 +1,166 @@
+
+#include <iostream>
+#include <triton/api.hpp>
+#include <triton/x86Specifications.hpp>
+
+using namespace triton;
+using namespace triton::arch;
+
+struct op {
+  unsigned int    addr;
+  unsigned char*  inst;
+  unsigned int    size;
+};
+
+struct op traceA = {0x400000, (unsigned char *)"\x66\x0F\xA5\xC2",             4}; /* shld       dx,ax,cl                    */
+struct op traceB = {0x400000, (unsigned char *)"\x0F\xA5\xC2",             3}; /* shld       edx,eax,cl                    */
+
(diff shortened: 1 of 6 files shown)


src/libtriton/includes/pythonBindings.hpp
```diff
@@ -14,6 +14,8 @@
   #include <python2.7/Python.h>
   #include <python2.7/longintrepr.h>
 #elif _WIN32
+  #include <cmath>
+  #define _hypot hypot
```
JonathanSalwan: Why this include and define?

src/examples/cpp/shld_test.cpp
```diff
@@ -0,0 +1,166 @@
+
```
JonathanSalwan: Thanks to provided tests but i'm using this [file](https://github.com/JonathanSalwan/Triton/blob/master/src/samples/ir_test_suite/ir.c) for my semantics tests and also the `qemu-test-x86_64`.

src/libtriton/arch/x86/x86Semantics.cpp
```diff
@@ -10481,6 +10483,85 @@ namespace triton {
         this->controlFlow_s(inst);
       }
 
+      void x86Semantics::shld_s(triton::arch::Instruction& inst) {
```
JonathanSalwan: Thanks for this function. The semantics is good and all my tests were successfully executed =). The coding style is not like others functions but it's not a big deal I will modify the coding style later.

JonathanSalwan:
Merged with fix

JonathanSalwan:
I also pushed `shrd` to close #413.

src/libtriton/includes/pythonBindings.hpp
gyorokpeter: mingw is giving errors in python.h if I leave these out. I figured out that I need both of these lines to avoid the error (otherwise it complains about either missing hypot or _hypot).

JonathanSalwan: Ok. Can you open a new issue for this mingw error and do a PR with theae both lines?

PR #458, opened 2017-01-10
check if the instruction reads/writes a part of the given register

gyorokpeter:


JonathanSalwan:
What about that:

```patch
-             if (pair.first.getId() == target.getConstRegister().getId())
-                return true;
+             if (pair.first.getParent().getId() == target.getConstRegister().getParent().getId())
+               return true;
```

gyorokpeter:
If an instruction writes to AH, then we don't want to match if we are checking whether it writes AL. However we do want to match if we are checking whether it writes EAX.

JonathanSalwan:
Mmmh okay. But in your case you are thinking about 32-bits. It's not true for 64-bits. E.g:  if you provide `EAX` as `target` and `pair.first.getId()` points to `AX`. Then, you don't catch it, right?

JonathanSalwan:
I thing you should deal with subset and bitvector. Thus, it is going to be valid for multi-architectures.

gyorokpeter:
It could be done like this:
```c++
            if (pair.first.getParent().getId() == target.getConstRegister().getParent().getId()) {
              if (pair.first.getLow() <= target.getConstRegister().getLow() && target.getConstRegister().getLow() <= pair.first.getHigh()) return true;
              if (target.getConstRegister().getLow() <= pair.first.getLow() && pair.first.getLow() <= target.getConstRegister().getHigh()) return true;
```

JonathanSalwan:
Something like that yep. And this just shown me that `if (pair.first.getAddress() == target.getConstMemory().getAddress())` is false if there is an overlap of store/load.

gyorokpeter:
This works for memory access:
```c++
            auto &&m1 = pair.first;
            auto &&m2 = target.getConstMemory();
            if (m1.getAddress() <= m2.getAddress() && m2.getAddress() < m1.getAddress()+m1.getSize())
              return true;
            if (m2.getAddress() <= m1.getAddress() && m1.getAddress() < m2.getAddress()+m2.getSize())
              return true;
```

PR #464, opened 2017-01-14
add removeWrittenRegister

gyorokpeter:


JonathanSalwan:
Why do you need this feature? I don't know if I should accept this PR. `writtenRegisters`should never been false and this feature looks like a patch when it fails.

gyorokpeter:
This is not a fix, I just found this function useful when I want to override the result to ignore the writing of certain registers. Of course it's also possible by adding long conditionals whenever encountering that instruction, but preserving this state in the object is cleaner.

JonathanSalwan:
Ok. Can you do the same thing for all others attrs (load, store, readReg, readImm)?

Thanks

JonathanSalwan:
Can you do the PR on dev-next please?

gyorokpeter:
I made this on top of the capstone 4.0 branch. If I tried to merge into dev-next it would also merge the capstone-related changes.

JonathanSalwan:
got it, i will push the patch with your name as author into dev-next.

PR #467, opened 2017-01-14
Fix #462

### LINKED ISSUE #462: Python binding not work with virtualenv
Before installing Triton, I switch to python virtual environment.
After that, I try to `import triton` from my virtual environment. 
It succeed once. ( The same session I install triton )
After closed the terminal, switch to virtual environment, and try to `import triton` again.
It failed.
I am not sure it's triton's problem, or the way I install triton is wrong.
Thanks.

bananaappletw:
Fix #462
Previous warning disappear.


JonathanSalwan:
Thanks!

PR #489, opened 2017-02-23
Build improve

pbrunet:
It is currently WIP but it may give you ideas about what is going on.


pbrunet:
@JonathanSalwan Can be review. You could.should review it commits per commits as it will be smaller steps.

PR #501, opened 2017-03-22
Registers

pbrunet:


pbrunet:
This MR looks good, failing tests already happened before but I need to do these modification to keep moving on less side effect.
The only remaining problem is about snapshot but I may need your expertise about what should happen :s

PR #502, opened 2017-03-23
CMake env

55-AA:


CMakeLists.txt
```diff
@@ -112,9 +112,10 @@ if(PYTHON_BINDINGS)
       set(PYTHON_LIBRARIES "$ENV{PYTHON_LIBRARIES}")
     endif()
     if(NOT PYTHON_INCLUDE_DIRS AND NOT PYTHON_LIBRARIES)
-        find_package(PythonLibs 2.7 REQUIRED)
-      else()
+      find_package(PythonLibs 2.7 REQUIRED)
+      if(NOT PYTHON_INCLUDE_DIRS AND NOT PYTHON_LIBRARIES)
```
JonathanSalwan: `find_package` already raises an exception if it cannot find paths.

src/libtriton/CMakeLists.txt
```diff
@@ -254,7 +253,13 @@ include_directories("${Boost_INCLUDE_DIRS}")
 
 # Define library's properties
 add_library(${PROJECT_LIBTRITON} ${LIBTRITON_KIND_LINK} ${LIBTRITON_SOURCE_FILES} ${LIBTRITON_PYTHON_SOURCE_FILES})
-add_dependencies(${PROJECT_LIBTRITON} gen-syscall32 gen-syscall64)
+if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+  add_dependencies(${PROJECT_LIBTRITON} gen-syscall64, gen-syscall32)
+endif()
+if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+  add_dependencies(${PROJECT_LIBTRITON} gen-syscall32)
+endif()
```
JonathanSalwan: I think you're wrong here.

CMakeLists.txt
55-AA: If set PYTHON_INCLUDE_DIRS and PYTHON_LIBRARIES as the params of CMake, line 117 would raise an exception.

PR #507, opened 2017-03-29
Fix/mingw build

pbrunet:


gyorokpeter:
"plateform" should be "platform"

src/libtriton/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ endif()
 
 # Default flags
 # To fix old python version bug http://bugs.python.org/issue21958
-if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
+if(${CMAKE_SYSTEM_NAME} MATCHES "Windows" AND NOT "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DHAVE_ROUND /MT")
 endif()
```
55-AA: Quoted variables, such as  "${CMAKE_CXX_COMPILER_ID}" , will cause a waring in windows CMake  version 3.7.2. 
`CMake_ Warning (dev) at src/libtriton/CMakeLists.txt:21 (if):`
`  Policy CMP0054 is not set: Only interpret if() arguments as variables or keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy details.  Use the cmake_policy command to set the policy and suppress this warning.`

`  Quoted variables like "MSVC" will no longer be dereferenced when the policy is set to NEW.  Since the policy is not set the OLD behavior will be used.`

Line 11 add  -D_CRT_SECURE_NO_WARNINGS , to avoid a few waring, such as:
`abstractBinary.cpp(50): warning C4996: 'fopen': This function or variable may be unsafe. Consider using fopen_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.`

pbrunet: Thanks for the report, I didn't really understand this warning under windows, I will do the change.

About the CRT_SECURE, I would prefer keep the warning and fix them than silent them all as the concern may be real. You report here the fopen call, did you see any other? appveryor didn't report them AFAIK

55-AA: All are fopen, no other.

pbrunet:
Every issue should be fixed: @JonathanSalwan do you agree with it?

JonathanSalwan:
> Every issue should be fixed: @JonathanSalwan do you agree with it?

Agreed. Can you rebase from `dev-next`, then I will merge `dev-next` into `master`.

Issue #521, opened 2017-04-24
Support Python3?


bananaappletw:
Triton is awesome project.
Thanks for the great work!!!

Is there any plan to support `python3`?
If that, maybe I could also help.
Thanks.

JonathanSalwan:
I never tried but maybe it already supported? Did you tried?

bananaappletw:
@JonathanSalwan 
Code seems like hard-coded python version as `2.7`.
Not providing way to customize python version
https://github.com/JonathanSalwan/Triton/blob/master/CMakeLists.txt#L115
https://github.com/JonathanSalwan/Triton/blob/master/src/libtriton/CMakeLists.txt#L282
Or something am I missing?
If I was wrong, correct me.
Thank you.

JonathanSalwan:
Yeah it's hardcoded but if you remplace them does it work?

JonathanSalwan:
If it works we can easily add an option to the compile process.

bananaappletw:
@JonathanSalwan 
Seems not.
There are another places of code need to be modified.
![capture](https://cloud.githubusercontent.com/assets/6241350/25344844/f7a387a8-2945-11e7-84da-fad82ebc8e70.PNG)

JonathanSalwan:
=(. Well, if you manage to figure out these issues I will be happy to accept your pull request :)

bananaappletw:
@JonathanSalwan
Ok, I will try!
Thank you for kind reply.

JonathanSalwan:
> Ok, I will try!

Woot, thanks. Go ahead from the `dev-next` branch.

bananaappletw:
https://docs.python.org/3/howto/cporting.html
Seems like `python3` binding for c++ is completely different compared to `python2`.
It needs some work to do that.

bananaappletw:
Quote from https://docs.python.org/2/c-api/intro.html
```
Note Since Python may define some pre-processor definitions which affect the standard headers on some systems, you must include Python.h before any standard headers are included.
```
The python binding codes should put `#include <Python.h>` before other headers

RyanLee1119:
@JonathanSalwan It seems Python3 is a must for Windows is you want to use **Tigress_protection**, because the dependent **Arybo** cannot compile under Python 2.7 in Windows:

> with the exception of Python 2 not available under Windows 
https://pythonhosted.org/arybo/setup.html

Steps to reproduce in Windows:
1. Build Triton in Python2.7
[...]

PR #523, opened 2017-04-25
[Progress of support python3] Fix python binding code

bananaappletw:
[Progress of support python3] #521 
There are some strange reasons previous compile succeed.
After I change `<python2.7/Python.h>` to `<Python.h>`
Compile failed.

Quote from https://docs.python.org/2/c-api/intro.html
```
Note Since Python may define some pre-processor definitions which affect the standard headers on some systems, you must include Python.h before any standard headers are included.
```
The python binding codes should put `#include <Python.h>` before other headers

I move every python binding related code on top of header list.
I will keep going on support `python3` without breaking current `python2` support.

### DIFF
--- a/src/libtriton/bindings/python/init.cpp
+++ b/src/libtriton/bindings/python/init.cpp
@@ -5,11 +5,10 @@
 **  This program is under the terms of the BSD License.
 */
 
-#include <iostream>
-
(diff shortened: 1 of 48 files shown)


bananaappletw:
@JonathanSalwan 
I think it could be a start.

PR #526, opened 2017-04-26
Remove duplicate find python code

bananaappletw:
I think this line should be redundant, maybe.
I am not sure, I haven't read all of `cmake` related code.
If I am wrong, correct me.
Thanks.

### DIFF
--- a/src/libtriton/CMakeLists.txt
+++ b/src/libtriton/CMakeLists.txt
@@ -281,8 +281,6 @@ install (DIRECTORY ${CMAKE_BINARY_DIR}/src/libtriton/includes/triton DESTINATION
 
 # Install Python bindings
 if(PYTHON_BINDINGS)
-    find_package(PythonInterp 2.7 REQUIRED)
-
     add_custom_target(python-triton ALL
       COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/${CMAKE_SHARED_MODULE_PREFIX}triton${CMAKE_SHARED_LIBRARY_SUFFIX}" "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/triton${PYTHON_SUFFIX}"
       DEPENDS triton


JonathanSalwan:
cc @pbrunet

pbrunet:
I would say there are no duplications as we check here for the interpreter while previous check is about libpython but @bananaappletw looks to have a better knowledge that me on this kind of think. May be I am wrong?

bananaappletw:
@JonathanSalwan 
@pbrunet 
Sorry, my mistake.
I mistaken `PythonLibs` to `PythonInterp`.
This line can't be remove.
I think this line should move to `/CMakeLists.txt`.
Maybe for better consistency?

bananaappletw:
@pbrunet 
Also, I think https://github.com/JonathanSalwan/Triton/blob/master/src/libtriton/CMakeLists.txt#L290
should be written as `setup.py` for better structure?

bananaappletw:
I just try to write `setup.py` in `bindings/python` directory.
But it seems to be difficult, many dependency are missing.
The ideal situation is like I am at `bindings/python` directory.
Type `python setup.py install`.
Then `triton` package is installed.

pbrunet:
You may have a look at : https://github.com/pbrunet/Triton/commit/4cff44f0b7e3c808784fae0b694fb6969fec1dca

It works correctly on linux but, as I am not a windows expert, I didn't succeed to make it works on windows.

There are multiple issue : 
1. Python library should link with the same crt than cpython and python2.7 was compiled using MSVC 2008 on  windows which don't support modern c++. But we live dangerously and hope the CRT didn't change to much...
2. setup.py will create a triton.so which link with libtriton.so. It is ok on linux but with windows, you have to export symboles from libtriton.dll (__declspec(dllimport)) and here I didn't succeed to make it works because of multiple crt linkage.

bananaappletw:
@pbrunet 
Thanks for the great explanation.
I want to do the same thing as you did then.
Maybe I will meet those issues, too.
Since that's hard to do that.
Maybe, I make `triton` support `python3` first?
Could you also checkout [this](https://github.com/JonathanSalwan/Triton/pull/527)?
This pull request move python code to the same place, just for better management.
Thanks.

pbrunet:
I really hope you will succeed to do it. I guess you will need to search an option that ask to export every symbole as std::string for example have to be exported.

I think python3 support and setup.py are not related. So you can do it in any order. The main issue will be about the python capi which should be adapted for some corner cases and guarded by python version preprocessor option.

One of the main benefit from using python3 will be to use modern msvc compiler with the setup but for compatibility, python2.7 support should still be available.

bananaappletw:
@pbrunet 
I think I could refactor some `cmake` related code.
Then latter work should be easier to proceed.
Thanks for the information you provided.

bananaappletw:
Actually, refactor to `setup.py` is more complicated than I thought.
I tried in recent days and found `cmake` is much simpler to dealing with it.

PR #534, opened 2017-05-11
Tools

pbrunet:


pbrunet:
About clang-format, you may have a look at : https://github.com/johnmcfarlane/unformat

Issue #549, opened 2017-05-25
Triton Compile Errors Using USer Compiled libboost


Owlz:
So I've been trying to compile/install Triton on-top of libboost that was compiled by me (and my CI service codeship). A strange thing happens. If lib boost is installed using apt-get (like locally on my box), the compile of Triton works fine. However, if I compile libboost myself, then attempt to compile Triton with that version, it spits out strange re-define errors:

```
cmake -DCMAKE_INSTALL_PREFIX=/home/rof/larissa -DCAPSTONE_INCLUDE_DIR=/home/rof/larissa/lib/python2.7/site-packages/capstone/include -DCAPSTONE_LIBRARY=/home/rof/larissa/lib/python2.7/site-packages/capstone/lib/libcapstone.so -DBoost_INCLUDE_DIR=/home/rof/larissa/boost/include -DBoost_LIBRARY_DIR=/home/rof/larissa/boost/lib ..
CPATH=/usr/include:/home/rof/larissa/include:/home/rof/larissa/include:/home/rof/larissa/include:/home/rof/larissa/boost/include make -j32 install
In file included from /home/rof/larissa/include/python2.7/Python.h:8:0,
from /tmp/pip-Le91zT-build/lib/triton/src/libtriton/includes/triton/pythonBindings.hpp:12,
from /tmp/pip-Le91zT-build/lib/triton/src/libtriton/bindings/python/init.cpp:10:
/home/rof/larissa/include/python2.7/pyconfig.h:1190:0: error: "_POSIX_C_SOURCE" redefined [-Werror]
#define _POSIX_C_SOURCE 200112L
^
In file included from /usr/include/x86_64-linux-gnu/c++/4.8/bits/os_defines.h:39:0,
from /usr/include/x86_64-linux-gnu/c++/4.8/bits/c++config.h:426,
from /usr/include/c++/4.8/iostream:38,
from /tmp/pip-Le91zT-build/lib/triton/src/libtriton/bindings/python/init.cpp:8:
/usr/include/features.h:230:0: note: this is the location of the previous definition
# define _POSIX_C_SOURCE 200809L
^
In file included from /home/rof/larissa/include/python2.7/Python.h:8:0,
from /tmp/pip-Le91zT-build/lib/triton/src/libtriton/includes/triton/pythonBindings.hpp:12,
from /tmp/pip-Le91zT-build/lib/triton/src/libtriton/bindings/python/init.cpp:10:
[...]

PR #550, opened 2017-05-25
Python Header To The Top

Owlz:
Ref #549 -- I was able to get the compiler errors worked out. The core issue is that some systems the python header files define system values that conflict with global values in other header files. It turns out, the solution for this is to import the python header files first. The other core system includes actually check for these values already being set and do not attempt to set them again.

All the changes made here are to move the python header includes to the top. I can confirm that these changes allowed Triton to be built in my case.

### DIFF
--- a/src/libtriton/bindings/python/init.cpp
+++ b/src/libtriton/bindings/python/init.cpp
@@ -5,12 +5,11 @@
 **  This program is under the terms of the BSD License.
 */
 
-#include <iostream>
-
 #include <triton/pythonBindings.hpp>
 #include <triton/pythonXFunctions.hpp>
 #include <triton/pythonBindings.hpp>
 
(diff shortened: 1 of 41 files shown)


bananaappletw:
@Owlz Maybe you could checkout [this comment](https://github.com/JonathanSalwan/Triton/issues/521#issuecomment-296917594)
`Python` related work continues in `dev-pb` branch.
I already did that in `dev-pb` branch.
Thanks.

Owlz:
Hmm. Is that dev branch stable? Should it be used instead of main branch? It seems to be referenced a lot but is still marked as development, thus I tend to not use it.

JonathanSalwan:
@Owlz `dev-pb` will be the next new API.

PR #555, opened 2017-06-04
add instance methods callback support

kamou:
this PR just brings the ability to provide instance method callbacks to addCallback, without having to decorate a method as `@staticmethod`, and thus, avoids having to define global variables. 

```
addCallback(someClass.memoryCaching, GET_CONCRETE_MEMORY_VALUE)
```
where `someClass.memoryCaching` may access to self.

### DIFF
--- a/src/libtriton/bindings/python/modules/tritonCallbacks.cpp
+++ b/src/libtriton/bindings/python/modules/tritonCallbacks.cpp
@@ -752,18 +752,37 @@ namespace triton {
         if (mode == nullptr || (!PyLong_Check(mode) && !PyInt_Check(mode)))
           return PyErr_Format(PyExc_TypeError, "addCallback(): Expects a CALLBACK as second argument.");
 
+        PyObject* function_self = nullptr;
+        PyObject* func = nullptr;
+
+        if (PyMethod_Check(function)) {
+          function_self = PyMethod_GET_SELF(function);
+          func = PyMethod_GET_FUNCTION(function);
+          Py_INCREF(func);
+        } else
+          func = function;
+
         try {
           switch (static_cast<triton::callbacks::callback_e>(PyLong_AsUint32(mode))) {
 
             case callbacks::GET_CONCRETE_MEMORY_VALUE:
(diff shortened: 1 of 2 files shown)


kamou:
woopsy, I'm failing the second addcalback python test. let me try fixing this.

src/libtriton/bindings/python/modules/tritonCallbacks.cpp
```diff
@@ -773,38 +792,57 @@ namespace triton {
...
+              triton::api.addCallback(callbacks::getConcreteRegisterValueCallback([func, function_self](triton::arch::Register& reg) {
                 /********* Lambda *********/
                 /* Create function args */
-                PyObject* args = triton::bindings::python::xPyTuple_New(1);
-                PyTuple_SetItem(args, 0, triton::bindings::python::PyRegister(reg));
+                PyObject* args = nullptr;
+                PyObject * pyReg = triton::bindings::python::PyRegister(reg);
+                if (function_self) {
+                  args = triton::bindings::python::xPyTuple_New(2);
+                  Py_INCREF(function_self);
```
pbrunet: I guess if you incref before the call, you should decref after the call, no? Or maybe it is done by the callee?

src/testers/unittests/test_im_callback.py
```diff
@@ -0,0 +1,43 @@
...
+        self.assertTrue(self.flag)
+
+        self.flag = False
+        removeCallback(self.cb_flag, CALLBACK.GET_CONCRETE_REGISTER_VALUE)
+        processing(Instruction("\x48\x89\xd8")) # mov rax, rbx
+        self.assertFalse(self.flag)
+
+    def cb_flag(self, x):
+        self.flag = True
+
```
pbrunet: It could be great if you can add a test with the self object dying before the callback to make sure incref/decref are ok and will not make crash the program.

pbrunet: Thx for this contribution and sorry for the delay. It looks great to have this!!

JonathanSalwan:
Rebased and pushed with 8da8e70b73fc3257c5a078266494c89a24396ded

PR #564, opened 2017-06-25
Elaborated the documentation in taintEngine.cpp

bntejn:
Elaborated the explanation about the tagging policy and propagation rules. Also rewrote the whole subsections to make the article *flow* better.


JonathanSalwan:
Thanks!

PR #565, opened 2017-06-27
Fix segfault on OSX

### LINKED ISSUE #552: ONLY_ON_SYMBOLIZED and freed node

JonathanSalwan:
- Add more tests
- Fix segfault on OSX
- Fix #552
- Add testing with Linux clang 3.8 on Travis
- Fix memory leaks over the CPython API
- Fix some wrong behaviors related to implicit / explicit semantics
- Remove useless code

### DIFF
--- a/src/libtriton/arch/irBuilder.cpp
+++ b/src/libtriton/arch/irBuilder.cpp
@@ -102,6 +102,13 @@ namespace triton {
       /* Clear previous expressions if exist */
       inst.symbolicExpressions.clear();
 
+      /* Clear implicit and explicit semantics */
+      inst.getLoadAccess().clear();
+      inst.getReadRegisters().clear();
+      inst.getReadImmediates().clear();
+      inst.getStoreAccess().clear();
+      inst.getWrittenRegisters().clear();
+
       /* Update instruction address if undefined */
       if (!inst.getAddress())
         inst.setAddress(this->architecture->getConcreteRegisterValue(this->architecture->getParentRegister(ID_REG_IP)).convert_to<triton::uint64>());
@@ -139,19 +146,21 @@ namespace triton {
        * is enable we must compute semanitcs to spread the taint.
        */
       if (!this->symbolicEngine->isEnabled()) {
+        /* Clear memory operands */
+        this->collectNodes(uniqueNodes, inst.operands, false);
+
+        /* Clear implicit and explicit semantics */
+        loadAccess.clear();
+        readRegisters.clear();
+        readImmediates.clear();
+        storeAccess.clear();
+        writtenRegisters.clear();
+
+        /* Symbolic Expressions */
         this->removeSymbolicExpressions(inst, uniqueNodes);
+
+        /* Restore backup */
         *this->symbolicEngine = *this->backupSymbolicEngine;
-      }
-
-      // ----------------------------------------------------------------------
-
-      /*
-       * If the symbolic engine is defined to process symbolic
-       * execution only on tainted instructions, we delete all
-       * expressions untainted and their AST nodes.
-       */
-      if (this->modes.isModeEnabled(triton::modes::ONLY_ON_TAINTED) && !inst.isTainted()) {
-        this->removeSymbolicExpressions(inst, uniqueNodes);
       }
 
       // ----------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


src/testers/unittests/test_semantics.py
```diff
@@ -149,6 +149,27 @@ def emulate(self, pc):
 
             self.assertTrue(ret)
 
+            try:
```
pbrunet: Here you only check that str() call doesn't crash/raise exception?
May be you can add in a test_utils.py this as a : check_instruction_data_is_valid() or something like this?

src/libtriton/arch/register.cpp
```diff
@@ -110,14 +126,15 @@ namespace triton {
...
-      concreteValueDefined(false) {
+        concreteValue(0),
+        concreteValueDefined(false) {
     }
 
 
-    Register::Register(const RegisterSpec& spec, triton::uint512 concreteValue)
-      : RegisterSpec(spec) {
-      this->setConcreteValue(concreteValue);
+    Register::Register(const Register& other)
```
pbrunet: and here

src/libtriton/arch/register.cpp
```diff
@@ -75,13 +83,21 @@ namespace triton {
...
     }
 
 
-    bool RegisterSpec::operator!=(const RegisterSpec& reg1) const {
-      return !(*this == reg1);
+    void RegisterSpec::operator=(const RegisterSpec& other) {
+      BitsVector::operator=(other);
+      this->name    = other.name;
+      this->id      = other.id;
+      this->parent  = other.parent;
```
pbrunet: Same here

src/libtriton/arch/register.cpp
```diff
@@ -26,6 +26,14 @@ namespace triton {
     }
 
 
+    RegisterSpec::RegisterSpec(const RegisterSpec& other)
+      : BitsVector(other) {
+        this->name    = other.name;
+        this->id      = other.id;
+        this->parent  = other.parent;
+    }
```
pbrunet: You should not define this function as it is already generated by the compiler and here you only give a worse possible version (the compiler can do better), you disable the move constructor and add more code to maintains.

src/libtriton/arch/irBuilder.cpp
```diff
@@ -292,12 +242,57 @@ namespace triton {
...
+    void IrBuilder::collectUnsymbolizedNodes(T& items) const {
+      T newItems;
+
+      for (auto it = items.cbegin(); it != items.cend(); it++) {
+        if (std::get<1>(*it) && std::get<1>(*it)->isSymbolized() == true)
+          newItems.insert(*it);
+      }
+
+      items.clear();
+      items = newItems;
```
pbrunet: Here you could "move" the new container and it would be done in O(1) instead of O(n)

src/libtriton/arch/irBuilder.cpp
```diff
@@ -292,12 +242,57 @@ namespace triton {
 
 
     void IrBuilder::removeSymbolicExpressions(triton::arch::Instruction& inst, std::set<triton::ast::AbstractNode*>& uniqueNodes) {
-      for (auto it = inst.symbolicExpressions.begin(); it != inst.symbolicExpressions.end(); it++) {
+      for (auto it = inst.symbolicExpressions.cbegin(); it != inst.symbolicExpressions.cend(); it++) {
```
pbrunet: You could use autofor to make the constness more explicit:

```
for(auto const& SE: inst.symbolicExpressions) {
}
```

src/libtriton/arch/irBuilder.cpp
```diff
@@ -162,72 +171,26 @@ namespace triton {
...
 
-          else if (std::get<0>(*it).getLeaAst() == nullptr && std::get<1>(*it)->isSymbolized() == false)
-            // FIXME: Should extract nodes
-            storeAccess.erase(it++);
+        /* Clear implicit and explicit semantics - MEM */
+        this->collectUnsymbolizedNodes(storeAccess);
 
-          else if (std::get<0>(*it).getLeaAst()->isSymbolized() == false && std::get<1>(*it)->isSymbolized() == false)
-            // FIXME: Should extract nodes
-            storeAccess.erase(it++);
```
pbrunet: I don't understand what was done with Lea checks here. Can you help me?
BTW, the new code looks far more readable, it is great!!

Issue #566, opened 2017-07-02
Falling through BUG?


wuyihao14:
case DWORD_SIZE:
            /* In AMD64, if a reg32 is written, it clears the 32-bit MSB of the corresponding register (Thx Wisk!) */
            if (this->architecture->getArchitecture() == triton::arch::ARCH_X86_64) {
              finalExpr = triton::ast::zx(DWORD_SIZE_BIT, node);
              break;
            }

Should this [line](https://github.com/JonathanSalwan/Triton/blob/master/src/libtriton/engines/symbolic/symbolicEngine.cpp#L929) be put outside `if`? Compiler gives warning `fall-through`

And many other fall-through warnings:

    /tmp/Triton-master/Triton-master/src/libtriton/arch/x86/x86Semantics.cpp: In member function ‘void triton::arch::x86::x86Semantics::bswap_s(triton::arch::Instruction&)’:
    /tmp/Triton-master/Triton-master/src/libtriton/arch/x86/x86Semantics.cpp:2892:50: error: this statement may fall through [-Werror=implicit-fallthrough=]
                bytes.push_front(triton::ast::extract(39, 32, op1));
                                ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
    /tmp/Triton-master/Triton-master/src/libtriton/arch/x86/x86Semantics.cpp:2893:11: note: here
            case DWORD_SIZE:
            ^~~~
    /tmp/Triton-master/Triton-master/src/libtriton/arch/x86/x86Semantics.cpp:2895:50: error: this statement may fall through [-Werror=implicit-fallthrough=]
                bytes.push_front(triton::ast::extract(23, 16, op1));
                                ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
    /tmp/Triton-master/Triton-master/src/libtriton/arch/x86/x86Semantics.cpp:2896:11: note: here
            case WORD_SIZE:
            ^~~~
    /tmp/Triton-master/Triton-master/src/libtriton/arch/x86/x86Semantics.cpp: In member function ‘void triton::arch::x86::x86Semantics::paddb_s(triton::arch::Instruction&)’:
    /tmp/Triton-master/Triton-master/src/libtriton/arch/x86/x86Semantics.cpp:6971:48: error: this statement may fall through [-Werror=implicit-fallthrough=]
[...]

PR #567, opened 2017-07-02
Code style and AST Optimizations

JonathanSalwan:
- Use the keyword `this->`
- Don't extend if zero
- Don't extract if it's the full size
- #566: Remove redundant code


JonathanSalwan:
cc @pbrunet

pbrunet:
LGTM but It doesn't work on windows 32 :s

Issue #581, opened 2017-07-19
Triton to Z3 and AST improvements


JonathanSalwan:
Getting model from z3 is not taking so long. The big part (in term of time) is in the `getFullAst` and `Z3_parse_smtlib2_string()` functions.

JonathanSalwan:
* Tree to tree and not tree to string to tree (e.g: `Z3_parse_smtlib2_string`) (speedup++, RAM--)
* Remove the following nodes: `bvdecl`, `function`, `compound` and `assert`
* `node::isLogical()` returns true if it's a logical node (e.g: `equal`, `bvuge`, `land`, `lnot`, ...)
* Remove `Z3Result`
* Change `land(expr1, expr2)` to `land([expr1, expr2, expr3, ...])`
* Change `lor(expr1, expr2)` to `lor([expr1, expr2, expr3, ...])`

PR #582, opened 2017-07-20
From Triton AST to Z3 AST

JonathanSalwan:
PR for #581

* Tree to tree and not tree to string to tree (e.g: `Z3_parse_smtlib2_string`) (speedup++, RAM--)
* Remove the following nodes: `bvdecl`, `function`, `compound` and `assert`
* `node::isLogical()` returns true if it's a logical node (e.g: `equal`, `bvuge`, `land`, `lnot`, ...)
* Remove `Z3Result` (not used anymore)
* Change `land(expr1, expr2)` to `land([expr1, expr2, expr3, ...])`
* Change `lor(expr1, expr2)` to `lor([expr1, expr2, expr3, ...])`
* Remove `AstVisitor` (not used anymore)
* Add the `getConcreteSymbolicVariableValue` function


src/libtriton/includes/triton/ast.hpp
```diff
@@ -140,131 +144,98 @@ namespace triton {
...
-
     //! `(bvadd <expr1> <expr2>)` node
     class BvaddNode : public AbstractNode {
       public:
         BvaddNode(AbstractNode* expr1, AbstractNode* expr2);
-        virtual void init(void);
-        virtual void accept(AstVisitor& v);
-        virtual triton::uint512 hash(triton::uint32 deep) const;
+        void init(void);
+        triton::uint512 hash(triton::uint32 deep) const;
```
pbrunet: We could still use the override keyword.

src/libtriton/includes/triton/ast.hpp
```diff
@@ -284,211 +254,181 @@ namespace triton {
...
     //! `(concat <expr1> <expr2> ...)` node
     class ConcatNode : public AbstractNode {
       public:
         ConcatNode(AbstractNode* expr1, AbstractNode* expr2);
-        ConcatNode(std::vector<AbstractNode* > exprs, AstContext& ctxt);
-        ConcatNode(std::list<AbstractNode* > exprs, AstContext& ctxt);
-        virtual void init(void);
-        virtual void accept(AstVisitor& v);
-        virtual triton::uint512 hash(triton::uint32 deep) const;
+        ConcatNode(std::vector<AbstractNode*> exprs, AstContext& ctxt);
```
pbrunet: having a const& on the container should be better.

PR #585, opened 2017-07-29
Improve BitsVector Python bindings

JonathanSalwan:
* Improve code base
* Add `bv.getMaxValue()` into python bindings

### DIFF
--- a/src/libtriton/bindings/python/objects/pyBitsVector.cpp
+++ b/src/libtriton/bindings/python/objects/pyBitsVector.cpp
@@ -198,37 +198,17 @@ namespace triton {
       };
 
 
-      PyObject* PyBitsVector(const triton::arch::Immediate& imm) {
+      template PyObject* PyBitsVector(const triton::arch::Immediate& op);
+      template PyObject* PyBitsVector(const triton::arch::MemoryAccess& op);
+      template PyObject* PyBitsVector(const triton::arch::Register& op);
+      template <typename T>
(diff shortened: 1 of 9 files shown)


src/libtriton/bindings/python/objects/pyImmediate.cpp
```diff
@@ -111,7 +111,7 @@ namespace triton {
 
       static PyObject* Immediate_getBitvector(PyObject* self, PyObject* noarg) {
         try {
-          return PyBitvector(*PyImmediate_AsImmediate(self));
+          return PyBitsVector(*PyImmediate_AsImmediate(self));
```
pbrunet: As we only user getter, it is not really important but if we succeed to use inheritance with the CPython API, we could use it and keep the initial Immediate Object.

src/libtriton/bindings/python/objects/pyBitsVector.cpp
```diff
@@ -178,43 +198,37 @@ namespace triton {
...
+        if (object != NULL)
+          object->bv= new triton::arch::BitsVector(mem);
 
         return (PyObject*)object;
       }
 
 
-      PyObject* PyBitvector(const triton::arch::Register& reg) {
-        Bitvector_Object* object;
+      PyObject* PyBitsVector(const triton::arch::Register& reg) {
```
pbrunet: All of them look so close that I can't stop thinking about templates.

PR #595, opened 2017-08-22
Implemented a feature to optionally add a custom "tag" when tainting

bntejn:
Also implemented a Python wrapper & Python object (`TaintTag`) to deal with the tagging when using Pin. A user can create any kind of custom tag in the form of a Python dictionary, which propagates through the memory regions and registers along with existing *taint* represented as bits.


JonathanSalwan:
Please rebase from `dev-v0.6`.

PR #597, opened 2017-08-24
Pass Z3 header locations through CMake

awruef:
If Z3 is installed in a non-standard location, but its path is given to cmake (say via `CMAKE_PREFIX_PATH`) then configuration will succeed  but compilation will fail because the include paths for Z3 are not propagated to the right place. The right fix here might be to have a `TRITON_INCLUDES` variable that wraps up all the includes necessary to compile something that includes a Triton header, but this quick fixes it for me for now.

### DIFF
--- a/src/examples/cpp/CMakeLists.txt
+++ b/src/examples/cpp/CMakeLists.txt
@@ -1,3 +1,4 @@
+include_directories(${Z3_INCLUDE_DIR})
 enable_testing()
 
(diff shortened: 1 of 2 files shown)


JonathanSalwan:
is this behavior still present in the `dev-v0.6`?

awruef:
Just checked, works fine there. Sorry for the bother!

PR #610, opened 2017-10-03
Remove some default destructors

JonathanSalwan:


pbrunet:
LGTM but it looks like there is an issue with the Semantic in travis' tests.

JonathanSalwan:
I really don't understand why travis failed. The fail is nondeterministic and Triton provides the correct concrete value...

For example:
```
[OK] 0x400a6d: mov rax, 0
[OK] 0x400a74: mov rbx, 0
[KO] 0x400a7b: tzcnt rbx, rax (2 register error(s))
     Register       : rbx
     Symbolic Value : 0000000000000040
     Concrete Value : 0000000000000000
```

`rbx` must be equal to 0x40. I don't know why the concrete value from Pin is 0.

PR #617, opened 2017-10-25
Fix Z3 include handling

marcusmueller:
It doesn't make sense to check for Z3_INCLUDE_DIRS being checked before calling
find_package(Z3) the first time.

Also, includes didn't work in the examples subfolder, if things weren't in the
default include dir right away.


JonathanSalwan:
Checks failed.

marcusmueller:
@JonathanSalwan as far as I can see, Travis fails because apt-get failed (which isn't really my patch's fault, is it ;) ?).

Got to admit I haven't used AppVeyor before and between the last time I used MSVC or CMake under Windows to build anything and now a decade has passed, so I'm still trying to grok that.

Could you re-run Travis?

marcusmueller:
Anyway, you're right, the AppVeyor test case highlights a shortcoming; I've got a fix coming up.

JonathanSalwan:
Erf, you have to rebase from `dev-v0.6`, I didn't saw you was on `master`.

PR #629, opened 2017-12-01
Allow debug build from CMake

thorstent:
This change makes Release the default mode if nothing is specified. You can however, make a debug build by specifying 

`cmake ..-DCMAKE_BUILD_TYPE=Debug`

The CMake mode will automatically set the correct -O


CMakeLists.txt
```diff
@@ -65,7 +65,16 @@ else()
     )
 endif()
 
-set(CMAKE_BUILD_TYPE Release)
+set(default_build_type "Release")
+ 
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+  message(STATUS "Setting build type to '${default_build_type}' as none was specified.")
+  set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE
+      STRING "Choose the type of build." FORCE)
```
pbrunet: Do you really need to cache this value as CMAKE_BUILD_TYPE is already cached?

src/libtriton/CMakeLists.txt
```diff
@@ -21,9 +21,6 @@ endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR ${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
     # Flags
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror -Wno-unused-parameter -Wno-unknown-pragmas -Wno-mismatched-tags")
-    if(NOT GCOV)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3")
-    endif()
```
pbrunet: You don't want to optimize code? I guess you are right for debug build but you should set this flag for release build.

I guess it could be done with CMAKE_CXX_FLAGS_RELEASE

CMakeLists.txt
thorstent: I got this pattern from here: https://blog.kitware.com/cmake-and-the-default-build-type/
Kitware are the makers of CMake

src/libtriton/CMakeLists.txt
thorstent: CMake by default sets them by default to O3 in Release mode (at least on my system).

You can see that by typing `make VERBOSE=1`. On my system I get 

`[  6%] Building CXX object src/libtriton/CMakeFiles/triton.dir/api/api.cpp.o
cd /home/tarracht/source/Triton-release/src/libtriton && /usr/bin/c++    -I/home/tarracht/source/Triton/src/libtriton/includes -I/home/tarracht/source/Triton-release/src/libtriton/includes  -Wall -Wextra -Werror -Wno-unused-parameter -Wno-unknown-pragmas -Wno-mismatched-tags -fno-stack-protector -fomit-frame-pointer -fno-strict-aliasing -fPIC -std=c++11 -O3 -DNDEBUG   -o CMakeFiles/triton.dir/api/api.cpp.o -c /home/tarracht/source/Triton/src/libtriton/api/api.cpp`

JonathanSalwan:
Fixed with 35c8252e6b9315bcad1fb7e4ac906cce161e1662

PR #631, opened 2017-12-02
Fix Dockerfile

Manouchehri:
Main mirrors for Ubuntu 15.10 are no longer up, so we should switch to 16.04 (Xenial).

Tested locally and also builds fine on Docker Hub. 

[`docker pull thawsystems/triton`](https://hub.docker.com/r/thawsystems/triton/builds/)


JonathanSalwan:
Thanks :)

PR #640, opened 2018-01-05
Changed solverModel destructor

mxmssh:
solverModel destructor is not virtual anymore to be able to compile
Triton with Ponce.


mxmssh:
My mistake, sorry guys.

PR #649, opened 2018-02-05
Remove assumptions that the system python is python2

fmagin:
The calls to the unittests failed when "python" was a link to "python3" and not to "python2" and thus make check failed at least on an Arch Linux.

I only tested it without the PIN integration, but I don't think this anything.


pbrunet:
You may want to use : ${PYTHON_EXECUTABLE} which is the one detected from CMake with find_program

JonathanSalwan:
Thanks! Can you rebase these three commits to make only one? Then i will merge this PR into the `dev-v0.6` branch.

Issue #651, opened 2018-02-06
Get Z3 expression from the python side


thebabush:
Hi again,

I'm trying to get a Z3 expression out of a Triton AstNode. I know I can do it from C++, but for now I'm prototyping so I would like to keep as much as possible in python.

I couldn't find anything in the docs, so I went ahead and wrote [a small wrapper](https://github.com/kenoph/thelp) using ctypes (leveraging Z3 python bindings).

I tried to do everything in a custom method inside pyTritonContext, but there's some problem I can't figure out while creating a python `z3.Ast` object :/
It's kind of hacky because I'm building a ctype object from C++, but if you think it could be useful I can try to fix what's wrong and submit a pull request. Otherwise, I'll just stick to my ctypes wrapper.

Thanks

JonathanSalwan:
Could be interesting to have this kind of feature directly into the `libTriton` but i don't know how to do it due to the fact that `libz3` does not exposes its python bindings directly into their lib. I'm not really conformable with the idea to split the API into several files (`.so` and `.py`). I would like to keep everything into a `.so` which is more standalone. If you manage to integrate this method into the `libTriton`, i will be more than happy to merge it into `dev-v0.6`.

kenoph:
#653

PR #654, opened 2018-02-06
Dirty convertAstToZ3 (#651)

thebabush:
This a very hacky way to interface with z3's ctypes bindings. AFAIK the ctypes API is not exported in the C Python API, so there are two options:

1 - Do a bunch of black magic with ctypes in order to get a pointer to python's `z3.main_ctx().ctx`
2 - Create a lightweight ctypes wrapper + some glue python (this is the proper way, but requires external python code)
3 - Do even more black magic and use 2, but implement it inside libtriton itself. Very very very hacky.

The other hacky part is creating the `z3.Ast` object from C. This is the situation:

```python
# ptr is an address set from C/C++

class Ast(ctypes.c_void_p):
  def __init__(self, ast): self._as_parameter_ = ast
  def from_param(obj): return obj

ast = Ast(ptr) # Doesn't work (well, the code works but the instance gives all sorts of errors). I suspect this is because Ast doesn't call c_void_p.__init__ and so self.value is not set

# This works but it's hacky. Everything works even without calling Ast.__init__
ast = c_void_p(ptr)
ast.__class__ = Ast.__class__

# This also works, even hackier
# Casting directly c_void_p to Ast fails :/
ast = ctypes.cast(ctypes.pointer(ctypes.c_void_p(ptr)), ctypes.POINTER(Ast))
```

I went with the `__class__` option and coded that in C/C++ so there's no dependency on external .py files.

Let me know what you think. It's also my first time messing with native Z3/Python. I tried to make sure the reference counts are updated correctly but I wouldn't bet on it.

P.S.: Sorry for the second PR, I didn't choose the right branch.

P.P.S.: On top of all these things, `z3.so` used by triton and python must be the same of course. I think `pip install z3` installs a local version inside the venv. I manually deleted it, which is suboptimal. Unfortunately, I don't think there's an easy fix for that.


JonathanSalwan:
Merged with an up-to-date modification for `dev-v0.7`. If you are also able to make `z3ToTriton` it will be awesome!

Thanks

PR #658, opened 2018-02-14
Re-wrote tritonToZ3AST::convert to not use recursion (to prevent running out of stack space)

thomasdullien:
Hey there,

This is an early / slightly premature pull request, mostly meant as FYI -- some tests are still failing, see below.

During my testing, Triton ended up crashing in applying path constraints since the recursion level of tritonToZ3Ast::convert exceeded a depth of 7k. In order to make sure we do not run out of stack space again, this code rewrites the function to not use recursion and instead performs iterative postorder traversal and an auxiliary unordered_map to keep intermediate results.

The new code currently fails 3 of the 'make tests' - tests; it does seem to generate the same model outputs for my small examples.

The tests that fail are:
17 - PinToolCrackMeHashCollision (Failed) 
18 - PinToolInjectModelWithSnapShot (Failed) 
21 - PinToolPathConstraints (Failed)

I will try to figure out what is going wrong - any hints about what files I should look at are much appreciated.

There is still one TODO in the code - right now, the code keeps the contents of the unordered_map during the entire conversion process; it is possible that better memory consumption / fewer cache misses can be obtained by cleaning out the child entries after they have been used.

I will happily perform any other cleanups that may become necessary, and any hints on how to do more thorough testing are happily received :)

Cheers,
Thomas


thomasdullien:
I decided to perform a much more lightweight refactoring inspired by pbrunet's fork.

PR #664, opened 2018-03-01
Adapt code to compile with Visual Studio 2013

illera88:


JonathanSalwan:
fixed things to be rebased from `dev-v06` and been merged.

PR #668, opened 2018-03-21
Use boost bimap for speed

thebabush:
Hi,

so I was doing some basic profiling on my tool and turns out that `SymbolicEngine::removeSymbolicExpression()` was taking almost 50% of the execution time. That's because of the final for loop. I guess this is due to `ONLY_ON_SYMBOLIZED` /`ONLY_ON_TAINTED`.

So I basically dropped in boost's bimap and gained quite some speed. I haven't checked for memory usage so I don't know how much it increased, but you might be interested in this PR anyway.


JonathanSalwan:
Thanks for this improvement but tests must be ok before merge :)

kenoph:
Ok, I think I got the problem. I should have read the bimap documentation more carefully lol.

So the thing is that both keys must be unique in a bimap, while that's not true for std::map.

The original implementation of `SymbolicEngine::removeSymbolicExpression()` is:

```cpp
   408   ¦ ¦ void SymbolicEngine::removeSymbolicExpression(triton::usize symExprId) {                                                                                     
   409   ¦ ¦ ¦ std::map<triton::uint64, triton::usize>::iterator it;                                                                                                      
   410                                                                                                                                                                    
   411   ¦ ¦ ¦ if (this->symbolicExpressions.find(symExprId) != this->symbolicExpressions.end()) { 
            ...
            ...
   424   ¦ ¦ ¦ ¦ /* Concretize the memory if it exists */                                                                                                                 
   425   ¦ ¦ ¦ ¦ for (it = this->memoryReference.begin(); it != memoryReference.end(); it++) {                                                                            
   426   ¦ ¦ ¦ ¦ ¦ if (it->second == symExprId) {                                                                                                                         
   427   ¦ ¦ ¦ ¦ ¦ ¦ this->concretizeMemory(it->first);                                                                                                                   
   428   ¦ ¦ ¦ ¦ ¦ ¦ return;                                                                                                                                              
   429   ¦ ¦ ¦ ¦ ¦ }                                                                                                                                                      
   430   ¦ ¦ ¦ ¦ }                                                                                                                                                        
   431   ¦ ¦ ¦ }
```

Based on the early exit in the final loop, is it safe to assume that the mapping is always 1-to-1? If this is not the case, I'm either missing something or the logic might be wrong... probably the first, but I would like to have confirmation :)

So basically, can there be two memory locations with same symbolic expression id?

JonathanSalwan:
Hey, thanks for this PR. Unfortunately it makes conflicts with my new updates. However, I made a patch and commited with your name.

Issue #568, opened 2017-07-03
compiling c++ examples on windows


chq007:
Hi Triton team,

I compiled Triton on windows and everything goes well. I can successfully run the python examples without any problems. However,  when I tried to compile c++ examples from Triton project using Visual Studio 2015 and added triton.lib, I got many Link errors as follows:

    Error LNK2001: unresolved external symbol "public: void __cdecl triton::API::setArchitecture(unsigned int)" (?setArchitecture@API@triton@@QEAAXI@Z)
    ErrorLNK2001	unresolved external symbol "class triton::API triton::api" (?api@triton@@3VAPI@1@A)	
	

I also tried to compile the c++ example using vs 2015 native tools command prompt using the following command and still got the same problem:

    cl ir.cpp /I path_to_triton\src\libtriton\includes /I path_to_boost\boost_1_64_0 /I path_to_z3\include triton.lib libz3.lib triton.lib

Would you please tell me how to compile the cpp examples?

JonathanSalwan:
I'm not really comfortable with the Windows OS. Maybe someone else may help?

JonathanSalwan:
Is it ok? Can I close it?

douggard:
I'm having the same problem with `triton.lib` trying to integrate it into a project. Comparing it (in a hex editor) with `xed.lib`, it is very short. It appears there are no entries for functions.

It also doesn't open as an AR archive (in 7z), which I guess is what `lib` files are. I will post any updates I figure out.

douggard:
There is a single export in the lib and the DLL, `inittriton` which seems related to the Python API. I don't really see any references to that in the C++ API.

douggard:
I think the code needs a bunch of `__declspec(dllexport)` to export functions on Windows. The reason the `inittriton` was exported was the `PyMODINIT_FUNC` is defined to that on Windows.

[An example PyMODINIT_FUNC definition.](https://github.com/python/cpython/blob/c06c22e9a9fb9326e79fcf1551601eacc1fd457d/Include/pyport.h#L640)

Maybe a solution like below, then pollute the code with `EXPORT` on every API function. Or maybe a script to fix it up on Windows. 

```c
#if defined(WIN32)
#define EXPORT __declspec(dllexport) 
#else
#define EXPORT __attribute__((visibility("default"))) 
#endif
```

JonathanSalwan:
Is there a flag on windows which allows us to export all functions at the compile time?

[...]

PR #670, opened 2018-03-22
TRITON_EXPORT must be set to nothing in Windows if static lib created

illera88:
If libtriton is created as dll then it works but if you try to link your project against libtriton as library it will produce linker errors.

If in Windows linking against libtriton as library (.lib) then use `BUILDING_LIB` as preprocessor.

### DIFF
--- a/src/libtriton/includes/triton/dllexport.hpp
+++ b/src/libtriton/includes/triton/dllexport.hpp
@@ -15,6 +15,8 @@
     #else
       #define TRITON_EXPORT __declspec(dllexport) // Note: actually gcc seems to also supports this syntax.
     #endif
+  #elif BUILDING_LIB
+    #define TRITON_EXPORT
   #else
     #ifdef __GNUC__
       #define TRITON_EXPORT __attribute__ ((dllimport))


illera88:
Related #568

src/libtriton/includes/triton/dllexport.hpp
```diff
@@ -15,6 +15,8 @@
     #else
       #define TRITON_EXPORT __declspec(dllexport) // Note: actually gcc seems to also supports this syntax.
     #endif
+  #elif BUILDING_LIB
```
JonathanSalwan: Who defined `BUILDING_LIB`?

illera88: I have to define it in my app if I am importing a `.lib` instead of a `.dll` in Windows.

JonathanSalwan: shouldn't it be defined by the cmakelist ?

illera88: Mmmm... I guess it will work too if you define it in the CMakeList.txt. Give me a minute and I'll test it

illera88: The important part is defining it in the target app (the one who is importing the static library). there must be no prefix/decoration if importing from a static library so it should be:
`#define TRITON_EXPORT`

Anyway, I know that `__declspec(dllexport)` is necessary in Windows but, I am not sure that `__declspec(dllimport)` is so maybe the line:
`#define TRITON_EXPORT __declspec(dllimport) // Note: actually gcc seems to also supports this syntax.` can be replace by #define TRITON_EXPORT

illera88: that will work for static libraries. Not sure about using `libtriton` as a dll

illera88:
Indeed just removing `__declspec(dllimport)` works for static and dynamic libraries.
I'll close this PR and do a new one

PR #691, opened 2018-06-11
Create a real interface for SMT solver

### LINKED ISSUE #663: Create a real interface for SMT solver
Right now Triton uses Z3 as SMT solver but does not provides a real interface so users can plug-in any desired SMT solver (CVC4, raSAT, OpenSMT,...), send the formulas to external servers to be solved and return the solution or (why not) solve them by hand :)

To do so `solverEngine.cpp` should be modified to make it a real interface that could be substituted by any way of formula solving that the user may want. 

Cheers

illera88:
Fix #663 

- Added the API `setSolver` to set a custom SMT solver.
- I've created examples on how to use a custom Solver interface.
- I've removed a CMakeLists condition to build tests for Windows too.
- Restrict some flags to only apply to *NIX builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,6 +135,7 @@ endif()
 # FIXME: We should move the tracer away to not link with these libraries
 
 if(Z3_INTERFACE)
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DZ3_INTERFACE")
 	# Find Z3
 	if(NOT Z3_INCLUDE_DIRS)
 		set(Z3_INCLUDE_DIRS "$ENV{Z3_INCLUDE_DIRS}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ option(KERNEL4 "Pin will run on a Linux's kernel v4" ON)
 option(PINTOOL "Build Triton with the Pin tool as tracer" OFF)
 option(PYTHON_BINDINGS "Enable Python bindings into the libtriton" ON)
 option(STATICLIB "Build a static library" OFF)
+option(Z3_INTERFACE "Use Z3 as SMT solver" ON)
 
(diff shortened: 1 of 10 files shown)


illera88:
No idea why the tests are failing in Windows. 

I actually don't know how tests works.

can you bring some light here?

Thanks

illera88:
The problem is that the tests cannot load `triton.dll`. There are some options:
- Installing `triton.dll` into system32
- Moving `triton.dll` to every test directory
- Using `WORKING_DIRECTORY ` argument for `add_test` and set it to wherever `triton.dll` is. This looks the most elegant solution but I have tried with no luck using this line:

`add_test(TaintRegister taint_reg WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/src/libtriton/Release)`

Do you know what am I missing?

illera88:
I verified that the problem is that `libz3.dll`, `triton.dll` and `capstone.dll` are not installed so it fails. If I move those three dlls to the directory where the tests are it works. 

```
10>------ Build started: Project: check, Configuration: Release x64 ------
10>  Test project C:/Users/user/Documents/code/forks/Triton/build_x64
10>      Start 1: TaintRegister
10>  1/5 Test #1: TaintRegister ....................   Passed    0.03 sec
10>      Start 2: InfoRegister
10>  2/5 Test #2: InfoRegister .....................   Passed    0.49 sec
10>      Start 3: IR
10>  3/5 Test #3: IR ...............................   Passed    0.06 sec
10>      Start 4: Simplification
10>  4/5 Test #4: Simplification ...................   Passed    0.50 sec
10>      Start 5: Constraint
10>  5/5 Test #5: Constraint .......................   Passed    0.72 sec
10>  
10>  100% tests passed, 0 tests failed out of 5
```

illera88:
I added a commit to restore things as they were before not running tests on Windows systems.

Can you merge this please.

Cheers

PR #712, opened 2018-08-13
Add AST optimizations

SweetVishnya:
Support arithmetic optimizations during AST creation.


SweetVishnya:
FAIL: test_let (test_ast_conversion.TestAstConversion)

Unfortunately, I don't understand the meaning of this test and can't fix it.

JonathanSalwan:
Thanks, let me few weeks to review the code.

JonathanSalwan:
I've not really read these lines yet but I think (maybe i'm wrong) some of these optimizations break some symbolic reasoning. Then, I think that it could be better to add these kind of optimizations in rules that can be enabled or disabled by users. It's a nice PR but I still have to think about all of this :).

SweetVishnya:
I check if one of the operands is not symbolic. For instance, eax+0 is still eax.

PR #724, opened 2018-09-26
Print more debug info in SymbolicVariable names

SweetVishnya:


JonathanSalwan:
We cannot name a python variable nor a smt variable with `:`.

SweetVishnya:
Tried to fix that if I got you correct.

JonathanSalwan:
Sorry for the delay. See #763.

PR #727, opened 2018-10-04
Fix OS X build with CMake

mrexodia:


JonathanSalwan:
Thanks for this PR. It will be awesome if you can also update our `.travis.ym` to support last osx version :)

mrexodia:
I can do it, can you tell me how to run the tests so I can run those on travis too?

JonathanSalwan:
You just have to uncomment these [lines](https://github.com/JonathanSalwan/Triton/blob/master/.travis.yml#L115-L126) and make a PR. Then, Travis will automatically pull the PR and applies the tests. Months ago these lines worked fine but with a recent OSX it does not work and I'm not really comfortable with the OSX ecosystem :)

mrexodia:
I’ll try tomorrow :) likely it was just the nonstandard class std::hash<>
thing because those are ancient osx versions being targeted...

JonathanSalwan:
Thanks! I really appreciate =).

JonathanSalwan:
I let this PR open so that you can push fixes for the travis file in this branch and then I will merge everything :)

mrexodia:
I did some quick tests with Travis for OS X only and everything appears to work. If the build passes it is ready to merge! I also updated the installation instructions specifically for OS X with the same generic method I used on Travis to make it easier for users in the future.

mrexodia:
Older OSX versions are making trouble. I removed Sierra, because soon a new macOS version should be released anyway. Right now only High Sierra is tested (which is what most people would be running anyway). The issue appears to be that `z3` cannot be installed because Travis has `python` installed and z3 wants `python@2` and they have conflicting links. I could manually remove `python` and install `python@2`, but stuff like this will only get more unsupported as newer OS versions are released so I don't think it's a good idea to hack around that.

JonathanSalwan:
Thanks for these fixes. I'm ok to only keep the High Sierra version.

mrexodia:
Okay, should be good to merge then!

JonathanSalwan:
Merged! Thanks @mrexodia, it was a very useful PR =).

PR #746, opened 2018-12-23
Python autocomplete helper module

Toizi:
I have been wanting to learn Triton several times but found the lack of autocomplete a pretty big hindrance.
Therefore I created a script that parses the doxygen documentation and generates a python module with stubs (including type hints) to enable IDE autocompletion support.
Since I figured it might be helpful to others as well, here is a PR. Feel free to decline if you don't want to include it in the main repository.

Example from VS Code
autocomplete vs code

Using go to definition provides an easy way to see the overloads and type information
go to definition

The generation is optional and can be enabled in the cmake build by setting `PYTHON_BINDINGS_AUTOCOMPLETE=1` and should be python 2/3 compatible without any dependencies.

To further improve the experience, it would make sense to add constructor documentation to the doxygen docs, since right now these accept any arguments.


JonathanSalwan:
Thanks that's very nice. What I have to do to maintain this "script generation" up-to-date in the future? For example, I see that I have to keep up-to-date the list of new classes register? There is something else?

Toizi:
As long the documentation follows the same style the script should continue working even if you add new classes.
It just processes all of the files in `bindings/python/objects/*.cpp` and `bindings/python/namespaces/*.cpp`.
The only special case is the register namespace, since it includes the documentation through htmlinclude or something along those lines which requires the script to also parse the .spec file.

JonathanSalwan:
Got it. That's nice. Let me few days to review it but I will probably merged it :).

 Thanks!

Toizi:
Sounds good. I refactored the code a bit and the script only generates a single `__init__.py` file now to reduce the extra o_ClassName from the import

PR #761, opened 2019-02-08
Makes newIstance call itself for new node's children. Closes #718

### LINKED ISSUE #718: triton::ast::newInstance doesn't make deep copy of AST
Actually I almost fixed it by my only and could make pull request. I only need to ask if Is it suitable solution:
* turn new node creations into `newNode = std::make_shared<SOMENODE>(*reinterpret_cast<SOMENODE*>(node, ctxt)); break;`
* Add `using AbstractNode:AbstractNode;` to every trivial derived class
* Implement custom constructor for nontrivial derived class (like `DecimalNode`, `ConcatNode` etc)

igogo-x86:
Don't see a better way to create deep copy.
Also I think constructor `AbstractNode::AbstractNode(const AbstractNode& other, AstContext& ctxt): ctxt(ctxt)` should be deleted as there's no way to use it. Can update this pull request if you agree with it

### DIFF
--- a/src/libtriton/ast/ast.cpp
+++ b/src/libtriton/ast/ast.cpp
@@ -30,18 +30,6 @@ namespace triton {
     }
 
 
-    AbstractNode::AbstractNode(const AbstractNode& other, AstContext& ctxt): ctxt(ctxt) {
-      this->eval        = other.eval;
-      this->parents     = other.parents;
-      this->size        = other.size;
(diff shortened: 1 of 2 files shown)


JonathanSalwan:
Thanks for this sweet fix.

> Can update this pull request if you agree with it

I'm agree with it.

PR #784, opened 2019-05-19
replace CMAKE_SOURCE_DIR with TRITON_ROOT to enable add_subdirectory usage

Toizi:
This change allows tool developers to build against triton through an `add_subdirectory` call, e.g.
```
add_subdirectory(extern/Triton)
target_link_libraries(myTool PRIVATE Triton::triton)
```
The alias `Triton::triton` has been introduced to allow library consumers to easily switch between `add_subdirectory` and `find_package` without changing the target to link against.
For libraries, the usage of `CMAKE_SOURCE_DIR` and `CMAKE_BINARY_DIR` should be discouraged when referring to the library source root.


CMakeLists.txt
```diff
@@ -38,7 +40,7 @@ if(PYTHON_BINDINGS)
     endif()
     find_package(PythonInterp ${PYTHON_VERSION} REQUIRED)
     add_custom_target(test-python
-        COMMAND ${PYTHONPATH_CMD} ${PYTHON_EXECUTABLE} -m unittest discover ${CMAKE_SOURCE_DIR}/src/testers/unittests
+        COMMAND ${PYTHONPATH_CMD} ${PYTHON_EXECUTABLE} -m unittest discover ${CMAKE_CURRENT_LIST_DIR}/src/testers/unittests
```
JonathanSalwan: Why not using the new variable `TRITON_ROOT`?

src/testers/CMakeLists.txt
```diff
@@ -5,27 +5,27 @@ if(PINTOOL AND ${CMAKE_SYSTEM_NAME} MATCHES "Linux" AND NOT PYTHON36)
...
-    add_test(PinToolTraceInst                   sh -c "${TRITON_SCRIPT} ${CMAKE_SOURCE_DIR}/src/examples/pin/trace_inst.py ${CMAKE_SOURCE_DIR}/src/samples/crackmes/crackme_xor a | tee /tmp/tester_triton_xxx && grep -nr syscall /tmp/tester_triton_xxx") # WIN OR FAIL, DUNNO
+    add_test(PinToolCheckUnsupportedSemantic    sh -c "${TRITON_SCRIPT} ${CMAKE_CURRENT_SOURCE_DIR}/unsuported_semantics.py ${TRITON_ROOT}/src/samples/ir_test_suite/ir | tee /tmp/tester_triton_xxx && ! grep -nr KO /tmp/tester_triton_xxx")
+    add_test(PinToolCheckSemantic               sh -c "${TRITON_SCRIPT} ${CMAKE_CURRENT_SOURCE_DIR}/check_semantics.py ${TRITON_ROOT}/src/samples/ir_test_suite/ir | tee /tmp/tester_triton_xxx && ! grep -nr KO /tmp/tester_triton_xxx")
+    add_test(PinToolCheckSemanticQemu           sh -c "${TRITON_SCRIPT} ${CMAKE_CURRENT_SOURCE_DIR}/qemu-test-x86_64.py ${TRITON_ROOT}/src/samples/ir_test_suite/qemu-test-x86_64| tee /tmp/tester_triton_xxx && ! grep -nr KO /tmp/tester_triton_xxx")
```
JonathanSalwan: Why not using `TRITON_ROOT` too?

Toizi: Do you mean instead of `CMAKE_CURRENT_SOURCE_DIR`?
It should really only be used to refer to directories outside of the current directory, since otherwise unnecessary dependencies on the directory structure are introduced.
But I think replacing these uses with `CMAKE_CURRENT_LIST_DIR` would be better

CMakeLists.txt
Toizi: I guess that is mostly a stylistic question since they both refer to the same thing in this case but I can change them if you prefer

Issue #783, opened 2019-05-06
setcc and cmovcc taint rules


Toizi:
### Behavior
I was experimenting with tainting and came across a piece of code similar to
```
cmp ecx (tainted), edx
setne dl
```
After the compare, zf is tainted, regardless of whether it is 0 or 1.
I expected that the dl register should be tainted as well, since its only input zf is tainted.

However, the taint is only propagated to the dl register if zf is zero. Is this correct behavior?

### Test sample
```
#!/usr/bin/env python2
from __future__ import print_function
from binascii import unhexlify
from struct import unpack

from triton import *
try:
    from triton_autocomplete import *
except ImportError:
    pass

def print_regs(ctx, regs):
    for reg in regs:
        print("{}: {:#x}".format(reg, ctx.getConcreteRegisterValue(getattr(ctx.registers, reg))))

def test_cmp_setne(reg1, reg2):
    # cmp ecx, edx
    # setne dl
    opcodes = ["\x39\xD1", "\x0F\x95\xC2"]

    ctx = TritonContext()
    ctx.setArchitecture(ARCH.X86_64)
    ctx.setAstRepresentationMode(AST_REPRESENTATION.PYTHON)

    ctx.setConcreteRegisterValue(ctx.registers.ecx, reg1)
    ctx.setConcreteRegisterValue(ctx.registers.edx, reg2)
    ctx.setTaintRegister(ctx.registers.ecx, True)

    for opcode in opcodes:
        inst = Instruction()
        inst.setOpcode(opcode)
        inst.setAddress(0x10000)

        print("before")
        print_regs(ctx, ['ecx', 'edx', 'zf'])

        if ctx.processing(inst) == False:
            print("inst not supported")
        print("\nexecuted\n  {}".format(inst))

        print("\nafter")
        print_regs(ctx, ['dl', 'zf'])
        print("tainted regs:")
        print(ctx.getTaintedRegisters())
        print("")
    return ctx.isRegisterTainted(ctx.registers.dl)


def main():
    print('testing equal values')
    print('====================================')
    eq_taint  = test_cmp_setne(0x123, 0x123)
    print('testing NOT equal values')
    print('====================================')
    neq_taint = test_cmp_setne(0x123, 0x120)
    print('====================================')
    print('equal taint     = {}'.format(eq_taint))
    print('not equal taint = {}'.format(neq_taint))


if __name__ == '__main__':
    main()
```

### Output
```
./test_setne.py
testing equal values
====================================
[...]

PR #787, opened 2019-05-25
cmov expression is now tainted if relevant condition flags are tainted

Toizi:
changes behavior mentioned in https://github.com/JonathanSalwan/Triton/issues/783#issuecomment-491860758


JonathanSalwan:
Thanks

Issue #792, opened 2019-07-03
Bug in AST duplication


werew:
I would like to make a copy of an AST (so that I can perform edits without modifying the original one) and evaluate it giving different symbolic variables. However when I set the symbolic variables this are concertized correctly in the original AST but not in its copy or unrolled counterpart: 

```python
from triton import *

ctx = TritonContext()
ctx.setArchitecture(ARCH.X86_64)
ac = ctx.getAstContext()

var1 = ctx.newSymbolicVariable(64, 'var1')
var2 = ctx.newSymbolicVariable(64, 'var2')

ast_original  = ac.bvadd(ac.variable(var1), ac.variable(var2))
ast_duplicate = ac.duplicate(ast_original)
ast_unrolled  = ac.unrollAst(ast_original)

ctx.setConcreteVariableValue(var1, 4)
ctx.setConcreteVariableValue(var2, 2)

print('AST ORIGINAL:  {}'.format(ast_original.evaluate()))
print('AST DUPLICATE: {}'.format(ast_duplicate.evaluate()))
print('AST UNROLLED:  {}'.format(ast_unrolled.evaluate()))
```

Output:
```
AST ORIGINAL:  6
AST DUPLICATE: 0
AST UNROLLED:  0
```

Is this a Triton bug or it should be done differently ?

JonathanSalwan:
Thanks for reporting that. I've pushed a patch. Tell me if it's ok for your use case :).

werew:
Thanks for the patch, everything works correctly now :)

werew:
There is still an issue: references are not handled correctly. 
Note the following example, here `ast_unrolled` remains unchanged since all references are missing, however  `ast_original` is modified as a side effect of modifying `st_duplicate`:

```
from triton import *

ctx = TritonContext()
ctx.setArchitecture(ARCH.X86_64)
ac = ctx.getAstContext()

var1 = ctx.newSymbolicVariable(64, 'var1')
var2 = ctx.newSymbolicVariable(64, 'var2')
var3 = ctx.newSymbolicVariable(64, 'var3')

var1_node = ac.variable(var1)
var2_node = ac.variable(var2)
var3_node = ac.variable(var3)

sym_expr= ctx.newSymbolicExpression(ac.bvadd(var1_node, var2_node),'my symexpr')

ast_original  = ac.bvmul(ac.reference(sym_expr), var1_node)
ast_duplicate = ac.duplicate(ast_original)
[...]

PR #795, opened 2019-07-25
Fixed parent cleaning in ast::newInstance

xarkes:
Before this patch, it was clearing a local copy of the parents vector, so it had no effect.

I don't know if it's the best patch, but now calling `removeParent` on the node results in the parents being effectively removed from the node's parents vector.

### DIFF
--- a/src/libtriton/ast/ast.cpp
+++ b/src/libtriton/ast/ast.cpp
@@ -2470,7 +2470,13 @@ namespace triton {
         throw triton::exceptions::Ast("triton::ast::newInstance(): No enough memory.");
 
       /* Remove parents as this is a new node which has no connections with original AST */
-      newNode->getParents().clear();
+      if (node->getType() != VARIABLE_NODE) {
+        /* VARIABLE_NODE are not duplicated (see #792), so don't remove their parents */
+        auto parents = newNode->getParents();
+        for (auto & p : parents) {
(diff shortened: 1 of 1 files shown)


JonathanSalwan:
In order to succeed tests the following update of the patch must be applied:

 ```patch
diff --git a/src/libtriton/ast/ast.cpp b/src/libtriton/ast/ast.cpp
index 3df8808d..0d494bb1 100644
--- a/src/libtriton/ast/ast.cpp
+++ b/src/libtriton/ast/ast.cpp
@@ -2470,7 +2470,13 @@ namespace triton {
         throw triton::exceptions::Ast("triton::ast::newInstance(): No enough memory.");
 
       /* Remove parents as this is a new node which has no connections with original AST */
-      newNode->getParents().clear();
+      if (node->getType() != VARIABLE_NODE) {
+        /* VARIABLE_NODE are not duplicated (see #792), so don't remove their parents */
+        auto parents = newNode->getParents();
+        for (auto& p : parents) {
+          newNode->removeParent(p.get());
+        }
+      }
 
       /* Create new instances of children and set their new parents */
       auto& children = newNode->getChildren();
```

Can you update your PR so that I can merge it?

xarkes:
Okay thanks, let's wait for the CI and see... :)

Issue #789, opened 2019-05-29
convertRegisterToSymbolicVariable is not working inside a GET_CONCRETE_REGISTER_VALUE callback


werew:
When trying to create a symbolic variable of a register using `convertRegisterToSymbolicVariable` inside a `GET_CONCRETE_REGISTER_VALUE` callback, the symbolic variable gets later overwritten.

Following is an example: here every time a memory location or a register is read we translate it into a symbolic variable.  The executed code is a simple addition or `rax` and the value at address `0x100`, both of which are read and translated to symbolic variables. 

The result is stored in `rax` which AST should look like this `(bvadd SymVar_1 SymVar_0)` where `SymVar_1` is the symbolic variable of `rax`. However `rax` is somehow concretized and we get `(bvadd (_ bv0 64) SymVar_0)`.

Notice that using `convertMemoryToSymbolicVariable` inside a `GET_CONCRETE_MEMORY_VALUE` callback works correctly.

```python
from triton import *
 
code = [
    b"\x48\x8b\x1d\x00\x01\x00\x00", # mov rbx, [0x100]
    b"\x48\x01\xd8", # add rax, rbx
]
 
mem_symvars = []
reg_symvars = []
seen_regs   = []
seen_mem    = []
 
########### CALLBACKS  ##############

def handle_mem_read(ctx, ma):
    global mem_symvars
    global seen_mem
 
    # Keep track of seen registers to avoid recursion
    addr = ma.getAddress()
    if addr in seen_mem: return
    seen_mem.append(addr)
 
    # Create symbolic var
    comment = "mem_{:#x}".format(ma.getAddress())
    symvar = ctx.convertMemoryToSymbolicVariable(ma,comment)
    mem_symvars.append(symvar)
 
def handle_reg_read(ctx, reg):
    global reg_symvars
    global seen_regs
 
    # Keep track of seen registers to avoid recursion
    reg_id = reg.getId()
    if reg_id in seen_regs: return
    seen_regs.append(reg_id)
 
    # Create symbolic var
    comment = "sym_reg_{}".format(reg.getName())
    symvar = ctx.convertRegisterToSymbolicVariable(reg, comment)
    reg_symvars.append(symvar)
 

############## RUN #################
 
# This function emulates the code.
def run(ip, ctx):
    for opcode in code:
        inst = Instruction()
        inst.setOpcode(opcode)
        inst.setAddress(ip)
        ctx.processing(inst)
 
        # Update IP
[...]

PR #797, opened 2019-08-05
Fix symbolization during callbacks

werew:
Fix for issue #789.

### DIFF
--- a/src/libtriton/engines/symbolic/symbolicEngine.cpp
+++ b/src/libtriton/engines/symbolic/symbolicEngine.cpp
@@ -740,6 +740,11 @@ namespace triton {
         /* Otherwise, use the concerte value */
         else {
           node = this->astCtxt->bv(this->architecture->getConcreteRegisterValue(reg), bvSize);
+
+          /* Double check the node is still not symbolic (eg. has been made symbolic during a callback) */
+          if (const SharedSymbolicExpression& symReg = this->getSymbolicRegister(reg)) {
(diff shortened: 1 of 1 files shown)


JonathanSalwan:
I'm not comfortable with the idea of "double check". What about this one?

```patch
diff --git a/src/libtriton/engines/symbolic/symbolicEngine.cpp b/src/libtriton/engines/symbolic/symbolicEngine.cpp
index f1c94dee..befd1879 100644
--- a/src/libtriton/engines/symbolic/symbolicEngine.cpp
+++ b/src/libtriton/engines/symbolic/symbolicEngine.cpp
@@ -732,6 +732,7 @@ namespace triton {
         triton::uint32 bvSize                = reg.getBitSize();
         triton::uint32 high                  = reg.getHigh();
         triton::uint32 low                   = reg.getLow();
+        triton::uint512 value                = this->architecture->getConcreteRegisterValue(reg);
 
         /* Check if the register is already symbolic */
         if (const SharedSymbolicExpression& symReg = this->getSymbolicRegister(reg)) {
@@ -739,7 +740,7 @@ namespace triton {
         }
         /* Otherwise, use the concerte value */
         else {
-          node = this->astCtxt->bv(this->architecture->getConcreteRegisterValue(reg), bvSize);
+          node = this->astCtxt->bv(value, bvSize);
         }
 
         /* extend AST if it's a extend operand (mainly used for AArch64) */
```

JonathanSalwan:
Merged

werew:
Nice, looks like a cleaner solution :)
Thanks for the fix!

PR #809, opened 2019-10-11
various fixes and improvements

PixelRick:


JonathanSalwan:
Thanks for this MR! Let me few days to review it.

src/libtriton/bindings/python/objects/pyTritonContext.cpp
```diff
@@ -2049,8 +2057,8 @@ namespace triton {
         }
 
         /* Free cb if exists */
-        if (PyMethod_Check(function))
-          Py_DECREF(cb);
+        /*if (PyMethod_Check(function))
+          Py_DECREF(cb);*/
```
JonathanSalwan: Can you explain this?

src/libtriton/bindings/python/objects/pyTritonContext.cpp
```diff
@@ -647,18 +603,7 @@ namespace triton {
...
-
-                  /* Fetch the last exception */
-                  PyErr_Fetch(&type, &value, &traceback);
-
-                  std::string str = PyStr_AsString(PyObject_Str(value));
-                  Py_XDECREF(type);
-                  Py_XDECREF(value);
-                  Py_XDECREF(traceback);
-                  throw triton::exceptions::Callbacks(str);
+                  throw triton::exceptions::PyCallbacks();
```
JonathanSalwan: Can you explain this?

src/libtriton/bindings/python/objects/pyTritonContext.cpp
```diff
@@ -677,6 +622,9 @@ namespace triton {
               return PyErr_Format(PyExc_TypeError, "Callbacks::addCallback(): Invalid kind of callback.");
           }
         }
+        catch (const triton::exceptions::PyCallbacks&) {
+          return nullptr;
+        }
```
JonathanSalwan: Can you explain this?

src/libtriton/engines/symbolic/symbolicExpression.cpp
```diff
@@ -155,11 +155,27 @@ namespace triton {
...
+
+        if ( !old
+          || (old->isSymbolized() != ast->isSymbolized())
+          || (old->evaluate() != ast->evaluate())
+          || (old->getBitvectorSize() != ast->getBitvectorSize())
+          || (old->isLogical() != ast->isLogical())
+          ) {
+          this->ast->initParents();
+        }
+
```
JonathanSalwan: I think it's more clear to let `this->ast->init();` For example, if in the future i'm adding new attributes in an AbstractNode, i have to remember to also add this attribute in this condition. So, I prefer to keep `ast->init()` :).

JonathanSalwan:
Also, i think there is missing the Python bindings for `lxor(expr1, expr2)` and `lxor([expr1, expr2, expr3])` in `pyAstContext.py`?

src/libtriton/arch/x86/x86Semantics.cpp
```diff
@@ -3325,6 +3325,12 @@ namespace triton {
         /* Spread taint */
         expr->isTainted = this->taintEngine->taintAssignment(src, src);
 
+        /* Tag undefined flags */
```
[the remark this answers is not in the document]
PixelRick: You mean to replace with "Tag undefined registers" ?

src/libtriton/includes/triton/astEnums.hpp
```diff
@@ -73,6 +73,7 @@ namespace triton {
       LET_NODE = 193,                 /*!< (let ((x y)) z) */
       LNOT_NODE = 197,                /*!< (and x y) */
       LOR_NODE = 199,                 /*!< (or x y) */
+      LXOR_NODE = 201,                /*!< (xor x y) */
```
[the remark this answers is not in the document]
PixelRick: Moved it to the end of the enum with value 239.

src/libtriton/ast/astContext.cpp
```diff
@@ -682,6 +682,13 @@ namespace triton {
     template TRITON_EXPORT SharedAbstractNode AstContext::lor(const std::vector<SharedAbstractNode>& exprs);
     template TRITON_EXPORT SharedAbstractNode AstContext::lor(const std::list<SharedAbstractNode>& exprs);
 
+    SharedAbstractNode AstContext::lxor(const SharedAbstractNode& expr1, const SharedAbstractNode& expr2) {
+      SharedAbstractNode node = std::make_shared<LxorNode>(expr1, expr2);
+      if (node == nullptr)
+        throw triton::exceptions::Ast("Node builders - Not enough memory");
+      node->init();
+      return node;
+    }
```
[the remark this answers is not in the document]
PixelRick: Added

src/libtriton/bindings/python/objects/pySymbolicVariable.cpp
```diff
@@ -215,6 +215,53 @@ namespace triton {
...
       }
 
+      static long SymbolicVariable_hash(PyObject* self) {
+        try {
+          return PySymbolicVariable_AsSymbolicVariable(self)->getId();
+        }
+        catch (const triton::exceptions::Exception& e) {
+          PyErr_Format(PyExc_TypeError, "%s", e.what());
+          return -1;
+        }
```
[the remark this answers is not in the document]
PixelRick: My bad, started with more complex code, this shouldn't throw.

src/libtriton/ast/ast.cpp
```diff
@@ -238,6 +243,7 @@ namespace triton {
       /* Init attributes */
       this->size = this->children[0]->getBitvectorSize();
       this->eval = ((this->children[0]->evaluate()) & this->getBitvectorMask());
+      this->symbolized = false;
```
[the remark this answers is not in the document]
PixelRick: init() is used as an update function when replacing children in the simplification passes. Without a reset to false it can stay true even if all children become non symbolized.
I found the bug when simplifying (VAR - VAR) by 0..

src/libtriton/bindings/python/objects/pyTritonContext.cpp
PixelRick: The python callbacks are wrapped into lambda functions in addCallback.
There is no increment of the reference there so it should not be decremented here.
To see the bug (python crash): add a python class's method as callback, then remove it explicitly, then re-add it.
removeAllCallbacks does not decrement any reference, so this one is safe.

The only solution I see to properly own a reference would be to add a unique_pointer with a deleter that calls Py_DECREF in the lambda capture..

src/libtriton/bindings/python/objects/pyTritonContext.cpp
PixelRick: If an exception happen in a python callback it is better to have the traceback than a single line error. The easiest way is to not format a new python error for this specific kind of exception and return null in the native python functions. I thus added specific catches for try clauses that do use callbacks directly or indirectly.

src/libtriton/bindings/python/objects/pyTritonContext.cpp
PixelRick: see above comment.

src/libtriton/engines/symbolic/symbolicExpression.cpp
PixelRick: 1/ why init and not initParents ?
2/ I agree it is clearly not the right place for this design-wise. However it is important when converting expressions to symbolic variables if they do already have huge parent trees. I added this when a scenario spent ~95% of the time in there.
I just added AstNode::hasSameConcreteValueAndType in in my next push. It would be interesting to use it in Simplification passes too to check that the given callbacks do give back proper nodes. When using multiple simplification callbacks, if one does change the value type it would only detected after the end of the processing loop when its new parent node initialize.
It can also be left undetected until late if the node is the one currently being assigned to an expression..

src/libtriton/arch/x86/x86Semantics.cpp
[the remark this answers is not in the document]
JonathanSalwan: yep

src/libtriton/ast/ast.cpp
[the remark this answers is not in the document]
JonathanSalwan: Oh nice catch! So, we have to add an test case in `./unittests/test_ast_simplification.py` to catch this bug in the future.

src/libtriton/bindings/python/objects/pyTritonContext.cpp
JonathanSalwan: Got it.

src/libtriton/bindings/python/objects/pyTritonContext.cpp
JonathanSalwan: Got it but still have another question :). Just to be sure, do you think we have to catch `PyCallback` also in others pyobjects?

src/libtriton/bindings/python/objects/pyTritonContext.cpp
JonathanSalwan: Can we add a test case of the crash in `unittests` and add a new comment at line 2060 in `pyTritonContext.cpp` like `FIXME: Add a unique_pointer with a deleter that calls Py_DECREF in the lambda capture.`

src/libtriton/engines/symbolic/symbolicExpression.cpp
JonathanSalwan: Got it but I still have to brainstorm about it. What I propose: Split this MR in two MRs. One without this feature so that I can merge all previous fixes and another one with this feature so that we can still discus about it and find a good way to handle this.

JonathanSalwan:
Review ok. What I propose:

1. Split this MR in two MR. One without modifying `symbolicExpression.cpp` and without adding `canReplaceNodeWithoutUpdate` and `hasSameConcreteValueAndTypeAs` so that I can merge everything else.
2. Open a new MR with feature about `canReplaceNodeWithoutUpdate` and  `hasSameConcreteValueAndTypeAs` so that we can discuss about it and find a proper way to handle this.

Thanks for these new features/fixes in Triton!

src/libtriton/bindings/python/objects/pyTritonContext.cpp
PixelRick: You mean in more places than I have added it or less ?
more: The PyCallbacks exception message is meant to inform of a missing catch: the catch blocks on general c++ exception I saw everywhere do raise a python error with the exception string.
less: Some catch blocks may be unnecessary, I initially put too much and ended up trimming the commit when composing the branch. Some indirect calls are less obvious than others, for instance convertRegisterToSymbolicVariable do call the simplification since it uses newSymbolicExpression.
If it needs to be as lean as possible, I'll check the whole list and add a comment with the call chain involved at each catch location.

src/libtriton/ast/ast.cpp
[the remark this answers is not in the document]
PixelRick: I am realizing something, that to fix the bug I just described processSimplification needs to be fixed too since it uses setChild without calling init !
(I have my own processSimplification which uses init but it is part of another PR).
This is another reason to have a canReplaceNodeWithoutUpdate.. since updating parent nodes for each simplified child node is nuts... So temporarily I will also add a check in the next commit to not update the parent if all children nodes have not been simplified (original shared_pointer == returned shared_pointer), is that ok ?

src/libtriton/engines/symbolic/symbolicExpression.cpp
JonathanSalwan: It's ok, i don't care about the history, you can rebase and force push.

src/libtriton/ast/ast.cpp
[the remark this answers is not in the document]
JonathanSalwan: ok. Do it into the second PR then.

PixelRick:
I added the unittest.
The vanilla post-bug state is kinda weird (when it does not simply crash completly) since the self.method_name returns None while the deletion happens on its im_func not the method object itself.
So I chose to compare the refcnt on a local ref to __func__, this way it does not crash inside removeCallback and the runtime seems to survive the wrong refcnt.

JonathanSalwan:
Hey @PixelRick,

Can you provide an PoC where I can reach PyCallbacks?

PixelRick:
`python
def raising_func():
    raise ValueError

def indirectly_raising_callback(ctx, _):
    raising_func()

def poc():
    ctxt = triton.TritonContext()
    ctxt.setArchitecture(triton.ARCH.X86_64)
    ctxt.addCallback(indirectly_raising_callback, triton.CALLBACK.GET_CONCRETE_REGISTER_VALUE)
    insn = triton.Instruction(b"\x48\x89\xd8")  # mov rax, rbx
    ctxt.disassembly(insn)
    ctxt.buildSemantics(insn)`

`Traceback (most recent call last):
  File "D:/projects/poc_cb/main.py", line 61, in <module>
    main()
  File "D:/projects/poc_cb/main.py", line 42, in main
    poc()
  File "D:/projects/poc_cb/main.py", line 32, in poc
    ctxt.buildSemantics(insn)
  File "D:/projects/poc_cb/main.py", line 23, in indirectly_raising_callback
    raising_func()
  File "D:/projects/poc_cb/main.py", line 19, in raising_func
    raise ValueError
ValueError`

PR #811, opened 2019-10-16
add header dependency to improve cmake add_subdirectory support

Toizi:
When linking against triton through cmake's `add_subdirectory()`, the header include paths were not automatically set when linking with `target_link_libraries(myllib PRIVATE Triton::triton)`

I have attached a small project to test this (includes readme):
[Triton_add_subdir_test.tar.gz](https://github.com/JonathanSalwan/Triton/files/3733301/Triton_add_subdir_test.tar.gz)

### DIFF
--- a/src/libtriton/CMakeLists.txt
+++ b/src/libtriton/CMakeLists.txt
@@ -57,8 +57,8 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux" OR ${CMAKE_SYSTEM_NAME} MATCHES "Darwin"
 endif()
 
 # Add Triton includes
-include_directories("${TRITON_ROOT}/src/libtriton/includes")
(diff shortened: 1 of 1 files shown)


JonathanSalwan:
Thanks @Toizi for this MR. If i'm keeping `TRITON_ROOT` and `CMAKE_BINARY_DIR` does it will work? The important point here is only `target_include_directories` right?

Toizi:
@JonathanSalwan the `TRITON_ROOT` could stay the same but `CMAKE_BINARY_DIR` has to be changed. It refers to the top level directory of the build (basically the same reason `TRITON_ROOT` is needed instead of `CMAKE_SOURCE_DIR`) and will be wrong when triton is added through `add_subdirectory`

PR #815, opened 2019-10-21
fix AST simplification pass not updating nodes' special attributes (e.g. "symbolized")

PixelRick:


src/libtriton/engines/symbolic/symbolicSimplification.cpp
```diff
@@ -189,15 +189,19 @@ namespace triton {
...
               /* Don't apply simplification on nodes like String, Integer, etc. */
               if (child->getBitvectorSize()) {
                 auto schild = this->callbacks->processCallbacks(triton::callbacks::SYMBOLIC_SIMPLIFICATION, child);
                 ast->setChild(index, schild);
+                needs_update |= !schild->canReplaceNodeWithoutUpdate(child);
                 worklist.push_back(schild);
               }
             }
+            if (needs_update)
+              ast->init();
```
JonathanSalwan: I think with the up-to-date `dev-v0.8`, here it's:

```cpp
ast->init();
ast->initParents();
```

So, I will update the prototype of the `AbstractNode::init()` function as follow `init(bool withParents=false)` and for this case, the line 204 will be changer by `ast->init(true);`

PR #820, opened 2019-10-25
Increase speed of nodes nodesExtraction algorithm and apply it in newInstance

igogo-x86:
This version of `nodesExtraction` works 10 times faster than current implementation (See N=20000 at test_ast_deep.py). Also I incorporated the usage of this function in `newInstance` that makes a deep copy of AST node.

I don't see a nice way to make added test fail with specified timeout. Seems like unittests doesn't support it. So this test probably doesn't test a lot (just works for very long time if speed degrades) so may be it should be deleted.

There's still exists the same problem when ascending AST(which is actually a DAG) traversal is happening in `AbstractNode::init` and `AbstractNode::initParents` functions. Current version will be calling the same methods several times for already processed elements and it never finishes when applying triton to real-live obfuscators. In my private version I comment them out but I would like to make the similar to nodesExtraction mechanism. So is it okey if I remove calling `initParent` from all `init` methods and make `initParent` first collecting all depending nodes and applying `init` to them?

### DIFF
--- a/src/libtriton/ast/ast.cpp
+++ b/src/libtriton/ast/ast.cpp
@@ -2492,16 +2492,20 @@ namespace triton {
       return newNode;
     }
 
-
     SharedAbstractNode unrollAst(const triton::ast::SharedAbstractNode& node) {
       return triton::ast::newInstance(node.get(), true);
     }
 
-
-    void nodesExtraction(std::deque<SharedAbstractNode>* output, const SharedAbstractNode& node, bool unroll, bool revert) {
-      std::unordered_map<triton::usize, std::set<SharedAbstractNode>> sortedlist;
-      std::deque<std::pair<SharedAbstractNode,triton::usize>> worklist;
-      triton::usize depth = 0;
+    /* Returns a vector of unique AST-nodes sorted topologically
+     *
+     * The result has property that any parent has smaller index than its children.
+     * @unroll - traverses through ReferenceNodes
+     * @revert - reverses the result
+     */
+    std::vector<SharedAbstractNode> nodesExtraction(const SharedAbstractNode& node, bool unroll, bool revert) {
+      std::vector<SharedAbstractNode> result;
+      std::set<AbstractNode *> visited;
+      std::stack<std::pair<SharedAbstractNode, bool>> worklist;
 
       if (node == nullptr)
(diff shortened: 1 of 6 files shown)


JonathanSalwan:
> I don't see a nice way to make added test fail with specified timeout. Seems like unittests doesn't support it. So this test probably doesn't test a lot (just works for very long time if speed degrades) so may be it should be deleted.

Yes, maybe we can remove this test.

> There's still exists the same problem when ascending AST(which is actually a DAG) traversal is happening in AbstractNode::init and AbstractNode::initParents functions. Current version will be calling the same methods several times for already processed elements and it never finishes when applying triton to real-live obfuscators. In my private version I comment them out but I would like to make the similar to nodesExtraction mechanism. So is it okey if I remove calling initParent from all init methods and make initParent first collecting all depending nodes and applying init to them?

If this is better for performance, yes go ahead! I'm always happy to merge MR which help us to scale on real world programs :)

JonathanSalwan:
mmmh, in fact let's keep this test.

JonathanSalwan:
> There's still exists the same problem when ascending AST(which is actually a DAG) traversal is happening in AbstractNode::init and AbstractNode::initParents functions. Current version will be calling the same methods several times for already processed elements and it never finishes when applying triton to real-live obfuscators. In my private version I comment them out but I would like to make the similar to nodesExtraction mechanism. So is it okey if I remove calling initParent from all init methods and make initParent first collecting all depending nodes and applying init to them?

btw, check out this MR #815 which is, I think, similar to the issue.

igogo-x86:
> mmmh, in fact let's keep this test.

Internet says there's https://pypi.org/project/timeout-decorator/ such a module to make tests with timeout, but I think it's better to avoid unnecessary dependencies.

> btw, check out this MR #815 which is, I think, similar to the issue.

Seems like my commits doesn't interact with it. Am I correct that `processSimplification` is called only for new nodes? So they don't have parent yet and no need to call `initParents` for simplified node

igogo-x86:
Some tests won't pass but looks like it's current branch problems

JonathanSalwan:
> Some tests won't pass but looks like it's current branch problems

I'm not sure this is due to the current branch.

JonathanSalwan:
I've created a branch (`dev-820`) to test this MR. If you can pull it and see what's happen it could be great. I got a crash in the `nodesExtraction` function.

Issue #753, opened 2019-01-18
shared_ptr and stack recursion on destructor calls


JonathanSalwan:

illera88:
Any new on this?

0xAcid:
Apéro

JonathanSalwan:
With: `ulimit -s 1024`

```python
from triton import *

ctx = TritonContext(ARCH.X86_64)
for i in range(10000):
    ctx.processing(Instruction(b"\x48\xff\xc0")) # inc rax
ctx.processing(Instruction(b"\x48\xc7\xc0\x00\x00\x00\x00")) # mov rax, 0
```

JonathanSalwan:
The commit 69ed47d implies a slowdown of the execution. The commit e185197 fixes this slowdown.

Below the time of the unittest `src/testers/unittests/test_simulation.py` for several commits. As we can see, the commit 69ed47d increases the time of the execution and the commit e185197 restores the time of the execution but still keeps the role of the gc in order to fix the isse #753.

```
Mar 30, 2020 e1851972b8c0c6dc35607c60d4102b8b449db2da 34.640s (Fix #753: Only keep shared_ptr at deep step level)
Mar 29, 2020 09d045070bce3af2b1b8bfbcc07f588d053dfeed 54.742s (missing const)
Mar 27, 2020 69ed47d44386c97447860f214718aaa70a296324 58.377s (Fix #753: shared_ptr and stack recursion on destructor calls (missing…)
Mar 26, 2020 4a41acec4744bad5aa3c7d2f3f013e73eb592ec8 36.043s (Fix #865: Instruction(address, opcode, size))
Mar 17, 2020 93f28b0c40164ba33f60847cf33e5d89d0f4c18c 37.296s (Add a new examples about synthetizing obfuscated expressions)
Feb 27, 2020 30cbace692cc8bc42eea9d026b74e41a1977eb5d 37.844s (Merge branch 'dev-v0.8' of ...)
Feb 08, 2020 fb3241e94a3e1d0be9831bfc7a865246ee4c9e30 37.600s (Fix Appveyor)
Dec 05, 2019 8f08fe61a73de30cdca0573d25a292357f480df6 34.217s (Fix include)
Oct 27, 2019 8cc318a7d13d921632d6acf43713fcd90afd4037 36.223s (Add missing initParents (#821))
[...]

PR #821, opened 2019-10-26
Remove parent initialization in LxorNode::init

igogo-x86:
Actually it was because I didn't have `LxorNode` in my branch and for this class initialization leads to infinite recursion

### DIFF
--- a/src/libtriton/ast/ast.cpp
+++ b/src/libtriton/ast/ast.cpp
@@ -2068,9 +2068,6 @@ namespace triton {
         if (this->children[index]->isLogical() == false)
           throw triton::exceptions::Ast("LxorNode::init(): Must take logical nodes as arguments.");
       }
-
-      /* Init parents */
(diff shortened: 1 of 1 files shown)


JonathanSalwan:
Ok, thanks. We just have to see now why it's not working on Windows

JonathanSalwan:
I think there is a stack overflow due to recursion somewhere.

```
$ ulimit -s 1024
$ make -C build/ check
..........................Segmentation fault
```

JonathanSalwan:
Ok, this is due to our `test_ast_deep.py` unittest and related to our https://github.com/JonathanSalwan/Triton/issues/753 bug.

JonathanSalwan:
It looks good now. There are still two missing initParents somewhere but don't worry I will add them and make unittest to catch the bug. Then, I will merge `dev-820` in `dev-v0.8`! Thanks for this MR!

PR #826, opened 2019-10-31
fixed xor taint propagation with same register + added unittest

### LINKED ISSUE #825: taint propagation wrong for "xor reg, reg"
As the title says, the taint propagation is wrong for `xor` if the source and destination register is the same. It should clear the taint since the result of the operation does not depend on the value of the tainted source/destination

Example:
```
eax is tainted
xor eax, eax  # this should clear the taint but it doesn't
eax is tainted
```

Toizi:
fixes #825

### DIFF
--- a/src/libtriton/arch/x86/x86Semantics.cpp
+++ b/src/libtriton/arch/x86/x86Semantics.cpp
@@ -13405,7 +13405,7 @@ namespace triton {
 
         /* Spread taint */
         /* clear taint if the registers are the same */
-        if (src.getRegister() == dst.getRegister())
+        if (dst.getType() == OP_REG && src.getRegister() == dst.getRegister())
           this->taintEngine->setTaint(src, false);
         else
           expr->isTainted = this->taintEngine->taintUnion(dst, src);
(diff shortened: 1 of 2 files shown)


JonathanSalwan:
This is false. Counter example `add rax, rax`.

Toizi:
@JonathanSalwan it doesn't touch `add` though. The if statement is in the `xor_s` function and not in a general taint spreading function

JonathanSalwan:
Arf, I didn't not saw this. I was thinking it was in the taintEngine.cpp. Sorry.

JonathanSalwan:
Let me think about this, because if we do that for the `xor_s` we have to do it for all others instructions... (see https://github.com/JonathanSalwan/Triton/issues/825#issuecomment-548381828)

JonathanSalwan:
btw, I think there is an error if we process the instruction `xor byte ptr [rax], 1`. In this case, I think `src.getRegister()` and `dst.getRegister()` are equal with an invalid register.

Toizi:
Good catch. I didn't realize you can xor memory directly but I'm not surprised since it's x86.

I think my latest commit should fix the issue

PR #854, opened 2020-02-27
Update Appveyor VS19 and support python 3

masthoon:
Hi, 

I removed the VS13/15/17 builds, updated to VS19 the python 2.7 builds and added python 3.6 builds.
Also, I bumped z3 version to 4.8.7, libboost to 1.71.0 and python 3 builds use LIEF version 0.10.1 (to run the tests).

It should produce 4 Triton builds with Visual Studio 2019:
- x86 .lib/.dll python27 .pyd
- x86 .lib/.dll python36 .pyd
- x64 .lib/.dll python27 .pyd
- x64 .lib/.dll python36 .pyd

I can restore the older VS builds if needed.

Thanks!

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,79 +12,55 @@ clone_folder: c:\projects\triton
 
 environment:
   CAPSTONE_INCLUDE_DIRS: c:\projects\triton\build\capstone-4.0.1\include\capstone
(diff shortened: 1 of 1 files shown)


JonathanSalwan:
This is awesome! Thanks!

PR #856, opened 2020-03-02
Fix update read registers

kamou:
Update instruction read registers when only one reference is found


JonathanSalwan:
Pushed with 6925d6012f2ff2c7c83a206ca3d0db09fd947618

PR #878, opened 2020-04-07
Add LGTM (Semmle) C++ analysis support

ekilmer:
Since libcapstone-dev is only v3.0.5 in Ubuntu 19.10, which is used by
LGTM to build Triton, we need to build Capstone v4.0.1 from source and
use it during the build.

Here are logs from this config to show that it works: https://lgtm.com/logs/47a1058f52b16ef96dbe50ca638d5e8bd9921d12/lang:cpp

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,10 @@ Full documentation is available on our [doxygen page](http://triton.quarkslab.co
   <a href="https://codecov.io/gh/JonathanSalwan/Triton">
     <img src="https://codecov.io/gh/JonathanSalwan/Triton/branch/dev-v0.7/graph/badge.svg" alt="Codecov" />
   </a>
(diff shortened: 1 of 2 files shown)


JonathanSalwan:
Sorry but I see no benefit to add this to Triton and to be honest it looks like more an ads for LGTM than a real contribution.

ekilmer:
Apologies for not providing enough context as to why I thought this might be useful, and it was not my intention to advertise for them. It's free for open source projects, and it (or someone else) had already tried to analyze Triton but failed during the C++ build without extra help. I can also totally understand not wanting to add another configuration file to your repo for something you don't personally want to use.

This project is very interesting and well-organized. Good luck!

PR #891, opened 2020-04-27
Add API method for pushing path condition object

apach301:
Add API method which allows you to push user-crafted pathCondition to pathManager.

### DIFF
--- a/src/examples/cpp/ctest_api.cpp
+++ b/src/examples/cpp/ctest_api.cpp
@@ -440,6 +440,60 @@ int test_8(void) {
 }
 
 
+int test_9(void) {
+  triton::API api;
+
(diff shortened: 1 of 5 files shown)


SweetVishnya:
@JonathanSalwan, we need it to add custom constraints to path predicate.

JonathanSalwan:
Thanks guys,

The patch looks good to me, I will merge it this week.

JonathanSalwan:
However, if you can add a test in this [file](https://github.com/JonathanSalwan/Triton/blob/dev-v0.9/src/examples/cpp/ctest_api.cpp#L467) to check that everything are working like you want, it will help :)

apach301:
@JonathanSalwan, added test_9 to check correctness of pushing custom path condition to the pathManager

PR #905, opened 2020-05-12
refactor x86 mul[x,u,s] semantic to avoir extract of [un]signed extension

RobinDavid:
I updated MUL, IMUL and MULX to avoid making extract of extension when operands and destination are of the same size. This does not break the semantic of formulas and **greatly** reduce sizes of ASTs.

I started writing few tests. One of them terminates with: `eflags: 0x46 (UC) != 0x42 (TT)` which is very weird as it seems to be the 3rd bit which change thus the parity flag. *(I am eager to have your idea on that)*.

Apart from that. It works like a charm.


src/libtriton/arch/x86/x86Semantics.cpp
```diff
@@ -5417,11 +5420,19 @@ namespace triton {
...
-            auto  expr = this->symbolicEngine->createSymbolicExpression(inst, this->astCtxt->extract(dst.getBitSize()-1, 0, node), dst, "IMUL operation");
+            auto  node1 = this->astCtxt->bvmul(op1, op2);
+            auto  node2 = this->astCtxt->bvmul(this->astCtxt->sx(dst.getBitSize(), op1), this->astCtxt->sx(src.getBitSize(), op2));
+            auto  node3 = (dst.getBitSize() == src.getBitSize()) ? node1 : this->astCtxt->extract(dst.getBitSize()-1, 0, node2);
+            // if(dst.getBitSize() == src.getBitSize()) {
+            //   auto expr = this->symbolicEngine->createSymbolicExpression(inst, node1, dst, "IMUL operation");
+            // }
+            // else {
+            //   auto expr = this->symbolicEngine->createSymbolicExpression(inst, this->astCtxt->extract(dst.getBitSize()-1, 0, node2), dst, "IMUL operation");
+            // }
```
JonathanSalwan, line 5431: Mmmh, why this condition (node3)? Can you provide an example where operands are of different size? Especially where the destination est smaller than the source (this is what you assume according to your extraction).

JonathanSalwan:
The PR looks good, i've only one question about the condition you added. Regarding PF, this flag in undefined, so it's ok if it does not match with Qemu state. However, it make me thought that we have to skip the verification of undefined flags in our unittest (I will do that).

src/libtriton/arch/x86/x86Semantics.cpp
RobinDavid: I just added ``node3`` to be consistent with all the other cases. But if it finds out src and dst are necessarily of the same size, the extract of sz is just useless as the semantic does not seems to keep the higher bits.

RobinDavid: I was not sure, that `src` and `dst` are necessarily of same size that's why I left the condition

RobinDavid:
I think I fucked up the pull request trying to rebase on dev-v0.9

src/libtriton/arch/x86/x86Semantics.cpp
RobinDavid: For this one the semantic is:

    ELSE IF (NumberOfOperands = 2)
        THEN
            TMP_XP ← DEST ∗ SRC (* Signed multiplication; TMP_XP is a signed integer at twice the width of the SRC *)
            DEST ← TruncateToOperandSize(TMP_XP);
            IF SignExtend(DEST) ≠ TMP_XP
                THEN CF ← 1; OF ← 1;
                ELSE CF ← 0; OF ← 0; FI;
Thus the higher part, is truncated and definitely lost. For operand sizes the manual specifies:

> Two-operand form — The source operand is promoted to 64 bits if it is a register or a memory location. The destination operand is promoted to 64 bits.

Thus I assume src and dst are necessarily of the same size. Thus we can in get rid of the `node3`. However `node2` is still required to compute cf and of.

PR #909, opened 2020-05-13
update semantic of imul, mul, mulx to avoid extract of sz/zx

RobinDavid:
Updated MUL, IMUL and MULX to avoid making extract of extension when operands and destination are of the same size. This does not break the semantic of formulas and greatly reduce sizes of ASTs.

For few cases I check that sizes of dst, src are the same before removing the extract/sx-zx, however I have not found examples were sizes were different.

### DIFF
--- a/src/libtriton/arch/x86/x86Semantics.cpp
+++ b/src/libtriton/arch/x86/x86Semantics.cpp
@@ -5365,13 +5365,14 @@ namespace triton {
                 auto dx    = triton::arch::OperandWrapper(this->architecture->getRegister(ID_REG_X86_DX));
                 auto op1   = this->symbolicEngine->getOperandAst(inst, ax);
                 auto op2   = this->symbolicEngine->getOperandAst(inst, src);
-                auto node  = this->astCtxt->bvmul(this->astCtxt->sx(triton::bitsize::word, op1), this->astCtxt->sx(triton::bitsize::word, op2));
-                auto expr1 = this->symbolicEngine->createSymbolicExpression(inst, this->astCtxt->extract(triton::bitsize::word-1, 0, node), ax, "IMUL operation");
-                auto expr2 = this->symbolicEngine->createSymbolicExpression(inst, this->astCtxt->extract(triton::bitsize::dword-1, triton::bitsize::word, node), dx, "IMUL operation");
+                auto node1 = this->astCtxt->bvmul(op1, op2);
+                auto node2 = this->astCtxt->bvmul(this->astCtxt->sx(triton::bitsize::word, op1), this->astCtxt->sx(triton::bitsize::word, op2));
+                auto expr1 = this->symbolicEngine->createSymbolicExpression(inst, node1, ax, "IMUL operation");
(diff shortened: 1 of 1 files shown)


src/libtriton/arch/x86/x86Semantics.cpp
```diff
@@ -5417,11 +5420,13 @@ namespace triton {
             auto& src  = inst.operands[1];
             auto  op1  = this->symbolicEngine->getOperandAst(inst, dst);
             auto  op2  = this->symbolicEngine->getOperandAst(inst, src);
-            auto  node = this->astCtxt->bvmul(this->astCtxt->sx(dst.getBitSize(), op1), this->astCtxt->sx(src.getBitSize(), op2));
-            auto  expr = this->symbolicEngine->createSymbolicExpression(inst, this->astCtxt->extract(dst.getBitSize()-1, 0, node), dst, "IMUL operation");
+            auto  node1 = this->astCtxt->bvmul(op1, op2);
+            auto  node2 = this->astCtxt->bvmul(this->astCtxt->sx(dst.getBitSize(), op1), this->astCtxt->sx(src.getBitSize(), op2));
+            auto  node3 = (dst.getBitSize() == src.getBitSize()) ? node1 : this->astCtxt->extract(dst.getBitSize()-1, 0, node2);
```
JonathanSalwan, line 5425: I think you can remove this condition (`node3` will always be `node1`). For the `IMUL` with two operands, `src` and `dst` are always of the same size and you cannot have a constant in `dst`.

* `0F AF /r | IMUL r16, r/m16`
* `0F AF /r | IMUL r32, r/m32`
* `REX.W + 0F AF /r | IMUL r64, r/m64`

src/libtriton/arch/x86/x86Semantics.cpp
```diff
@@ -5432,11 +5437,13 @@ namespace triton {
             auto& src2 = inst.operands[2];
             auto  op2  = this->symbolicEngine->getOperandAst(inst, src1);
             auto  op3  = this->symbolicEngine->getOperandAst(inst, src2);
-            auto  node = this->astCtxt->bvmul(this->astCtxt->sx(src1.getBitSize(), op2), this->astCtxt->sx(src2.getBitSize(), op3));
-            auto  expr = this->symbolicEngine->createSymbolicExpression(inst, this->astCtxt->extract(dst.getBitSize()-1, 0, node), dst, "IMUL operation");
+            auto  node1 = this->astCtxt->bvmul(op2, op3);
+            auto  node2 = this->astCtxt->bvmul(this->astCtxt->sx(src1.getBitSize(), op2), this->astCtxt->sx(src2.getBitSize(), op3));
+            auto  node3 = (dst.getBitSize() == (src1.getBitSize() == src2.getBitSize())) ? node1 : this->astCtxt->extract(dst.getBitSize()-1, 0, node2);
```
JonathanSalwan, line 5442: You can also remove this condition and use `node1`.

JonathanSalwan:
I will do the fix

JonathanSalwan:
Fixed. I'm curious about how much this PR improve performances if you did benchmark =).

RobinDavid:
Thank you for the quick merge !
I did! I will show you the results, its very satisfying :smile:

Issue #908, opened 2020-05-13
Improve the taint analysis


JonathanSalwan:
I don't like the current taint analysis, I will probably rewrite the taint engine. If you have some recommendation / discussion, feel free to comment this thread. Other threads taking into account:

* #473
* #827
* #888

SweetVishnya:
In #888 we decided to check whether instruction is symbolized like follows:

1. Get instruction operands from DynamoRIO.
2. Ask Triton if any registers or memory regions are symbolic.
3. If none of them are symbolized, then skip instruction processing (building AST).

But we didn't have time to try it out yet.

On the other hand, if one does not have a concolic engine, he should consider some mechanism to skip non-tainted instructions. Triton engine actually takes a lot time to build ASTs. IMHO, there should be a lightweight taint to select instructions for symbolic interpretation (reasoning/execution).

SweetVishnya:
As far as I see now, taint engine is useless if you don't have an emulation (concolic execution).

SweetVishnya:
What is your use case model for taint?

archercreat:
Ability to track multiple sources would be good.

SweetVishnya:
> Ability to track multiple sources would be good.

Do you want taint colors to distinguish multiple inputs?

archercreat:
> > Ability to track multiple sources would be good.
> 
> Do you want taint colors to distinguish multiple inputs?

Yes.

JonathanSalwan:
> What is your use case model for taint?

Actually, I never use the taint engine. I mainly use symbolic variable (we can consider that it's a kind of taint) and modes `ONLY_ON_SYMBOLIZED`, `ALIGNED_MEMORY`. Then, asking for a model we can deduct:

* No symbolic expression assigned to the reg/mem = untainted
* UNSAT = tainted but not controllable
* SAT = tainted and controllable

JonathanSalwan:
The taint analysis could be applied on a standalone static IR (#473)

PR #913, opened 2020-05-16
x86 flags semantics fix

Mahorori:
spread taint if symbolized and dont remove from written register

### DIFF
--- a/src/libtriton/arch/x86/x86Semantics.cpp
+++ b/src/libtriton/arch/x86/x86Semantics.cpp
@@ -1329,7 +1329,7 @@ namespace triton {
         /* Create the symbolic expression */
         auto expr = this->symbolicEngine->createSymbolicExpression(inst, node, cf.getConstRegister(), "Carry flag");
 
-        if (op2->evaluate()) {
+        if (op2->isSymbolized() || op2->evaluate()) {
           /* Spread the taint from the parent to the child */
           expr->isTainted = this->taintEngine->setTaintRegister(cf.getConstRegister(), parent->isTainted);
         }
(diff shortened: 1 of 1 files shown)


SweetVishnya:
Seems to be a part of #908.

JonathanSalwan:
> Seems to be a part of #908.

Yep, I will probably merge this PR but maybe this code will be removed if we rewrite the taint analysis.

JonathanSalwan:
Actually, there are some things not really natural for me in this MR:

- If we do that, we have to do the same thing for every conditional instructions.
- It's not really a dynamic taint analysis anymore (taint following the concrete execution).
- What if it's unsat?

SweetVishnya:
I personally don't like checking symbolic state in taint.

Mahorori:
if "x86_cl" register is symbolized and I do processing instruction "shl eax,cl"
I expect cf flag is in writtenRegisters or tainted.

if I understand correctly symbolized != concrete?
not sure about taint engine tbh so maybe it isn't great idea

JonathanSalwan:
`writtenRegisters` defines what registers are really written according to the concrete execution. It not assumes the possibilities of the symbolic reasoning. So, cf will be tainted or tagged as written register if and only if the concrete execution writes into cf.

Mahorori:
hm I see. I guess I had wrong idea of how triton works then.
Probably bad idea for taint analysis yet I think since symbolic expression contains IterNode and It doesn't hurt anything it's not bad to have those in writtenRegisters.
if that's not the way I'll close

JonathanSalwan:
I think we can close this pr. Sorry @Mahorori, but maybe the #908 will fix your issue.

PR #941, opened 2020-08-18
Add x86 VMOVD/VMOVQ semantics

SweetVishnya:
Added VMOVD/VMOVQ AVX instructions. Also, I noticed a strange thing -- MOVD and MOVQ are handled differently. MOVD XMM gets [zero extended](https://github.com/JonathanSalwan/Triton/blob/master/src/libtriton/arch/x86/x86Semantics.cpp#L6823) while MOVQ XMM remains [unchanged](https://github.com/JonathanSalwan/Triton/blob/master/src/libtriton/arch/x86/x86Semantics.cpp#L7323). Intel manual says:

**MOVD/Q with XMM destination:**

Moves a dword/qword integer from the source operand and stores it in the low 32/64-bits of the destination XMM register. The upper bits of the destination are zeroed. The source operand can be a 32/64-bit register or 32/64-bit memory location.

128-bit Legacy SSE version: Bits (MAX_VL-1:128) of the corresponding YMM destination register remain unchanged. Qword operation requires the use of REX.W=1.

VEX.128 encoded version: Bits (MAX_VL-1:128) of the destination register are zeroed. Qword operation requires the use of VEX.W=1.

EVEX.128 encoded version: Bits (MAX_VL-1:128) of the destination register are zeroed. Qword operation requires the use of EVEX.W=1.

So, what is the correct way of handling such instructions?

### DIFF
--- a/src/libtriton/arch/x86/x86Semantics.cpp
+++ b/src/libtriton/arch/x86/x86Semantics.cpp
@@ -320,8 +320,10 @@ UNPCKHPD                     | sse2       | Unpack and Interleave High Packed Do
 UNPCKHPS                     | sse1       | Unpack and Interleave High Packed Single-Precision Floating-Point Values
 UNPCKLPD                     | sse2       | Unpack and Interleave Low Packed Double-Precision Floating-Point Values
 UNPCKLPS                     | sse1       | Unpack and Interleave Low Packed Single-Precision Floating-Point Values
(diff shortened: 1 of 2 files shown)


JonathanSalwan:
> So, what is the correct way of handling such instructions?

mmmmh. I don't know yet. Have to think about it =/

PR #946, opened 2020-09-11
Docker: update Z3 to 4.8.9

XVilka:
Due to the SMT fuzzing efforts many bugs were fixed since 4.8.8.


JonathanSalwan:
PR, must be done on `dev-0.9`. Can you also update appveyor and travis, not only the Docker file?

XVilka:
Ah, right. Sorry, forgot about that.

PR #962, opened 2020-10-28
Concretize address in initLeaAst

### LINKED ISSUE #956: Support mode for getting concrete values for address expressions
Some day we should support computing memory access address from the concrete state, not from symbolic AST. Symbolic asts may have errors in concrete values due to undertaint or errors in symbolic engine. So, we should have a mode that will get memory expression address from the concrete state.

PaDarochek:
Closes #956

### DIFF
--- a/src/libtriton/engines/symbolic/symbolicEngine.cpp
+++ b/src/libtriton/engines/symbolic/symbolicEngine.cpp
@@ -1269,7 +1269,13 @@ namespace triton {
             if (mem.getPcRelative()) {
               baseValue = mem.getPcRelative();
             }
-            triton::uint64 concreteAddress = baseValue + indexValue * scaleValue + dispValue;
+            triton::uint64 concreteAddress = baseValue + dispValue;
+            if (index.isSubtracted()) {
+              concreteAddress -= indexValue * scaleValue;
+            }
+            else {
+              concreteAddress += indexValue * scaleValue;
+            }
 
             auto regSize = this->architecture->gprBitSize();
             if (regSize != 64) {
--- a/src/libtriton/engines/symbolic/symbolicEngine.cpp
+++ b/src/libtriton/engines/symbolic/symbolicEngine.cpp
@@ -1261,11 +1261,29 @@ namespace triton {
           mem.setLeaAst(leaAst);
 
           /* Initialize the address only if it is not already defined */
-          if (!mem.getAddress() || force)
-            mem.setAddress(leaAst->evaluate().convert_to<triton::uint64>());
+          if (!mem.getAddress() || force) {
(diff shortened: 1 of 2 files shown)


JonathanSalwan:
Can you explain in what scenarios `triton::modes::CONCRETIZE_ADDRESS` should be used instead of the AST evaluation ?

SweetVishnya:
AST evaluation may be wrong due to undertaint or missing instructions translations. We are getting real life examples for that. Computing memory address from concrete state allows us to avoid such errors. For instance:
```
mov eax, [ebx]
```
If ebx contains invalid AST, wrong memory will be dereferenced. While if we concretize ebx from concrete state, we dereference correct memory.

JonathanSalwan:
> We are getting real life examples for that.

Using a DBI or in pure emulation with Triton? (In pure emulation, concrete state are update from AST evaluation).

> Computing memory address from concrete state allows us to avoid such errors.

This is true only if you use a DBI to keep synch between the Triton state and the DBI state right?

Do you see any case where your patch must **not** be used as default behavior?

src/libtriton/engines/symbolic/symbolicEngine.cpp
```diff
@@ -1261,12 +1261,30 @@ namespace triton {
...
+              triton::uint64 indexValue = this->architecture->isRegisterValid(index) ? this->architecture->getConcreteRegisterValue(index).convert_to<triton::uint64>() : 0;
+
+              baseValue += segmentValue;
+              if (mem.getPcRelative()) {
+                baseValue = mem.getPcRelative();
+              }
+              triton::uint64 concreteAddress = baseValue + indexValue * scaleValue + dispValue;
+
+              auto regSize = this->architecture->gprBitSize();
+              if (regSize != 64) {
```
JonathanSalwan, line 1279: I think you can remove the `if` and directly perform the `&=` whatever the regSize is equal to 64 or not.

SweetVishnya:
Yeah, we use DBI, and this feature should be default. I updated the pull request.

src/libtriton/engines/symbolic/symbolicEngine.cpp
SweetVishnya: This is `-Wshift-count-overflow` and breaks my tests, because of `O3` optimizations.

SweetVishnya:
We solved this issue by synchronization of concrete and symbolic memory states. We compare concrete values received from DBI with Triton ASTs evaluation. If values are different, we concretize the corresponding register or memory. This also helps us to kill some symbolic registers after syscalls. So, now all memory accesses are correctly computed and there are no symbolic branches with wrong taken direction. @JonathanSalwan, you can close this PR and a corresponding issue #956.

PR #964, opened 2020-11-12
Add memory limit parameter for z3

apach301:


JonathanSalwan:
Thanks for this MR. Looks good to me. Let me few days before merging.

PR #970, opened 2020-11-25
Force to use python3 binary

XVilka:
Some distributions do not set `/usr/bin/python` if only python3 is installed, for example Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=970375

I tried to compile Triton in this environment (Debian Sid and python3-only) and it fails due to `/usr/bin/env: 'python': No such file or directory`. At this point it's better to use `python3` binary explicitly.

- https://wiki.debian.org/Python/2Removal
- https://access.redhat.com/solutions/4455511
- https://pythonclock.org/

### DIFF
--- a/src/examples/python/backward_slicing.py
+++ b/src/examples/python/backward_slicing.py
@@ -1,7 +1,7 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 ## -*- coding: utf-8 -*-
 ##
-## $ python ./backward_slicing.py
+## $ python3 ./backward_slicing.py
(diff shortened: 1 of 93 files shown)


SweetVishnya:
Hmm, interesting, I thought pin was working only with python 2.

JonathanSalwan:
> Hmm, interesting, I thought pin was working only with python 2.

Yes, @XVilka you have to keep `python2` for every Pin script (` src/examples/pin/*`)

XVilka:
Right, forgot about that 🤦

XVilka:
Updated the PR.

PR #971, opened 2020-11-25
Allow ONLY_ON_SYMBOLIZED clean to happen manually

jordan9001:
I need the speed and memory benefits of "ONLY_ON_SYMBOLIZED", but still would like to test some instruction semantics after processing an instruction.
This patch adds a mode "MANUAL_CLEAN_INSTRUCTION" which will delay doing most of what was in `IrBuilder::postIrInit` until the user manually calls `API::manualClearInstruction`

The use pattern is something like this
```
Triton.setMode(MODE.ONLY_ON_SYMBOLIZED, True)
Triton.setMode(MODE.MANUAL_CLEAN_INSTRUCTION, True)
while pc:
        # Fetch opcode
        opcode = Triton.getConcreteMemoryAreaValue(pc, 16)

        # Create the Triton instruction
        instruction = Instruction()
        instruction.setOpcode(opcode)
        instruction.setAddress(pc)

        # Process
        Triton.processing(instruction)

        ###
        # ... use instruction semantics here, like getLoadAccess or getWrittenRegisters
        ###

        Triton.manualClearInstruction(instruction)

        # Next
        pc = Triton.getConcreteRegisterValue(Triton.registers.rip)

```
This can be used just as easily with ONLY_ON_TAINTED as well, to check some information before you throw away untainted info.

If you would rather have this implemented some other way, let me know. I am happy to change things. This seemed to be the simplest method without any code duplication.

Thanks!

### DIFF
--- a/src/libtriton/api/api.cpp
+++ b/src/libtriton/api/api.cpp
@@ -515,6 +515,10 @@ namespace triton {
     this->arch.disassembly(inst);
   }
 
+  void API::manualClearInstruction(triton::arch::Instruction& inst) {
+    this->irBuilder->clearIrInst(inst);
+  }
+
 
 
   /* Processing API ================================================================================ */
--- a/src/libtriton/arch/irBuilder.cpp
+++ b/src/libtriton/arch/irBuilder.cpp
@@ -127,6 +127,16 @@ namespace triton {
 
 
     void IrBuilder::postIrInit(triton::arch::Instruction& inst) {
+      /* Set the taint */
+      inst.setTaint();
+
+      if (!this->modes->isModeEnabled(triton::modes::MANUAL_CLEAN_INSTRUCTION)) {
+        this->clearIrInst(inst);
+      }
+    }
+
+
+    void IrBuilder::clearIrInst(triton::arch::Instruction& inst) {
       std::vector<triton::engines::symbolic::SharedSymbolicExpression> newVector;
 
       auto& loadAccess        = inst.getLoadAccess();
@@ -134,9 +144,6 @@ namespace triton {
       auto& readImmediates    = inst.getReadImmediates();
       auto& storeAccess       = inst.getStoreAccess();
       auto& writtenRegisters  = inst.getWrittenRegisters();
-
-      /* Set the taint */
-      inst.setTaint();
 
       // ----------------------------------------------------------------------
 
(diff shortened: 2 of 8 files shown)


jordan9001:
I still have to add tests to make sure the new code passes the code coverage checks

JonathanSalwan:
I'm not sure this is the best way to handle it. I'm afraid that if we start adding modes for dirty hacks around here, around there it will be a mess very quickly and the user will do not understand what modes do under the wood... What about adding a new kind of callback like POST_IR and you do what you want inside the callback?

jordan9001:
I agree, that could be a lot cleaner.
So you would rather have an example look like this?

```
Triton.setMode(MODE.ONLY_ON_SYMBOLIZED, True)
insWasWrite = False
def postir_cb(triton_ctx, instruction):
        global insWasWrite
        insWasWrite = instruction.isMemoryWrite()
Triton.addCallback(CALLBACK.POST_IR, postir_cb)
while pc:
        # Fetch opcode
        opcode = Triton.getConcreteMemoryAreaValue(pc, 16)

        # Create the Triton instruction
        instruction = Instruction()
        instruction.setOpcode(opcode)
        instruction.setAddress(pc)

        # Process
        Triton.processing(instruction)

        if insWasWrite:
                ###
                # ...
                ###

        # Next
        pc = Triton.getConcreteRegisterValue(Triton.registers.rip)

```

JonathanSalwan:
> So you would rather have an example look like this? 

Yes for example. 

Now, when seeing why you need this, I'm wondering if we should keep `isMemoryWrite` still working when using `ONLY_ON_*`.

jordan9001:
I was only using isMemoryWrite as an example, but I do think it is one that lots of people could need.

My actual use case is to track memory addresses being dereferenced. I plan to take a instruction trace and do some reverse taint analysis, while still running with ONLY_ON_SYMBOLIZED.

JonathanSalwan:
Got it. For your use case using callbacks on GET/SET_CONCRETE_MEMORY_VALUE do not do the job?

jordan9001:
After thinking about it, yes, I think those callbacks would do the job for that issue!
I was thinking I would also need access to getLoadAccess and getStoreAccess. But I think I could make due with the callbacks.
Thanks!
If you still want a POST_IR callback, I am happy to help. Otherwise you can close this PR.

PR #975, opened 2020-12-02
Fix - clear assembly stringstream in Instruction

SweetVishnya:


JonathanSalwan:
Can you tell me why this is better to do this instead of doing the clear?

SweetVishnya:
Clear does not actually clear a stream, it only clears stream flags and assembly string still exists.

SweetVishnya:
https://stackoverflow.com/questions/20731/how-do-you-clear-a-stringstream-variable

PR #987, opened 2021-02-02
Utility functions

SweetVishnya:
The main purpose of this PR is supporting comments for path constraints. We need to set them for custom crafted path constraints.

### DIFF
--- a/src/libtriton/engines/symbolic/symbolicEngine.cpp
+++ b/src/libtriton/engines/symbolic/symbolicEngine.cpp
@@ -7,6 +7,7 @@
 
 #include <cstring>
 #include <new>
+#include <set>
 
 #include <triton/exceptions.hpp>
 #include <triton/coreUtils.hpp>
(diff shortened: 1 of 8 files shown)


SweetVishnya:
@JonathanSalwan, could you restart the pipeline, please? `get-pip.py` failed.

SweetVishnya:
Added a commit that removes redundant symbolic variables from SMT dump

src/libtriton/includes/triton/ast.hpp
```diff
@@ -822,6 +822,9 @@ namespace triton {
     //! Returns a deque of collected matched nodes via a depth-first pre order traversal.
     TRITON_EXPORT std::deque<SharedAbstractNode> search(const SharedAbstractNode& node, triton::ast::ast_e match=ANY_NODE);
 
+    //! Unroll references for a given AST.
+    TRITON_EXPORT SharedAbstractNode refUnroll(SharedAbstractNode node);
```
JonathanSalwan, line 826: Should not it be `SharedAbstractNode refUnroll(const SharedAbstractNode& node)`?

JonathanSalwan (commented):
I also have a doubt about the name `refUnroll`, I'm afraid that it can be weird to have a method called `unroll` and `refUnroll`. What do you think?

src/libtriton/engines/symbolic/symbolicEngine.cpp
```diff
@@ -451,47 +452,49 @@ namespace triton {
...
+        std::set<usize> symVars;
+        for (const auto& se : ssa) {
+          symExprs.push_back(se.first);
+          auto children = ast::childrenExtraction(se.second->getAst(), false, false);
+          for (const auto& child : children) {
+            if (child->getType() == ast::VARIABLE_NODE) {
+              auto id = reinterpret_cast<ast::VariableNode *>(child.get())->getSymbolicVariable()->getId();
+              symVars.insert(id);
+            }
+          }
```
JonathanSalwan, line 467: Is the code above only for searching symbolic variables? Why not use `ast::search(expr->getAst(), ast::VARIABLE_NODE)`?

SweetVishnya:
> I also have a doubt about the name `refUnroll`, I'm afraid that it can be weird to have a method called `unroll` and `refUnroll`. What do you think?

Can we use `dereference` then?

P.S. I will fix all review issues on Wednesday.

src/libtriton/includes/triton/ast.hpp
SweetVishnya: I pass it by value because I need to assign to shared_ptr later.

```
node = ref->getSymbolicExpression()->getAst();
```

src/libtriton/engines/symbolic/symbolicEngine.cpp
SweetVishnya: Done

src/libtriton/includes/triton/ast.hpp
JonathanSalwan: I think we can avoid this overhead by dealing with `AbstractNode*` instead of `SharedAbstractNode`. My point is that we can gain a speed up execution during the loop if we do not construct new `shared_ptr` at each iteration. What do you think about something like this:

```cpp
    SharedAbstractNode dereference(const SharedAbstractNode& node) {
      triton::ast::AbstractNode* ptr = node.get();

      while (ptr->getType() == REFERENCE_NODE) {
        const ReferenceNode* ref = reinterpret_cast<const ReferenceNode*>(node.get());
        ptr = ref->getSymbolicExpression()->getAst().get();
      }

      return ptr->shared_from_this();
    }
```

SweetVishnya: Ok, I'll fix it in next commit

SweetVishnya: Done

SweetVishnya:
I fixed all review issues

PR #997, opened 2021-03-02
Improve performance while setting concrete memory during initialization

ek0:
I was playing around with triton and big binaries (~50-200MB). I was noticing huge load times trying to concretize whole segments during initialization.

By avoiding iterating on the `alignedMemoryReference` map when empty, we save roughly 50% of CPU time. Trying to concretize 50MB on init went from ~60s to 30s.

This will still happen if you try to concretize big memory section after the symbolic execution engine did some calculation, as the map won't be empty. But this fix will allow users to optimize load times by concretizing memory during initialization at least.

### DIFF
--- a/src/libtriton/engines/symbolic/symbolicEngine.cpp
+++ b/src/libtriton/engines/symbolic/symbolicEngine.cpp
@@ -173,6 +173,12 @@ namespace triton {
 
       /* Removes an aligned memory */
       void SymbolicEngine::removeAlignedMemory(triton::uint64 address, triton::uint32 size) {
+        // Avoid accessing the alignedMemoryReference array when empty. This usually happens when
+        // you initialize the symbolic engine and concretize whole sections of an executable using
+        // setConcreteMemoryValue. No symbolic memory has been created yet but this function will
+        // still try to rougly erase (size * 7) elements.
+        if(this->alignedMemoryReference.empty())
+          return;
         /* Remove overloaded positive ranges */
         for (triton::uint32 index = 0; index < size; index++) {
           this->alignedMemoryReference.erase(std::make_pair(address+index, triton::size::byte));


SweetVishnya:
You can just use `setConcreteMemoryValue` from `Cpu`. That will just update the concrete state and won't try to concretize symbolic memory. `api->getCpuInstance()->setConcreteMemoryValue(...)`

https://github.com/JonathanSalwan/Triton/blob/master/src/libtriton/arch/x86/x86Cpu.cpp#L661

SweetVishnya:
@JonathanSalwan, one more thing I am wondering about whether `std::unordered_map` has the same speed as storing all concrete values by pages (like virtual memory emulation).

SweetVishnya:
Maybe, we can try storing mapping from page address to 4Kb buffer with a bitset of defined bytes?

JonathanSalwan:
When I have to deal with big binaries like yours, to speed up the execution when loading segments I setup [memory caching](https://github.com/JonathanSalwan/Triton/blob/master/src/examples/python/ctf-writeups/defcon-2016-baby-re/solve.py#L114-L120) areas. Then, I define a [callback](https://github.com/JonathanSalwan/Triton/blob/master/src/examples/python/ctf-writeups/defcon-2016-baby-re/solve.py#L271) on every concrete LOAD which will then use `setConcreteMemoryValue` of the [need memory cells](https://github.com/JonathanSalwan/Triton/blob/master/src/examples/python/ctf-writeups/defcon-2016-baby-re/solve.py#L241-L255). I other words, I use `setConcreteMemoryValue` on need memory cells and skip unneeded ones.

JonathanSalwan:
> Maybe, we can try storing mapping from page address to 4Kb buffer with a bitset of defined bytes?

I don't know what to think. In one hand I would like something simple and clear, in other hand it could nice to speed up the execution when initializing the memory.

PR #999, opened 2021-03-03
Enable to use Python 3.9

romainthomas:


SweetVishnya:
Can we specify python version as a CMake option? Having a separety option for each Python version seems not to be universal.

JonathanSalwan:
Yeah i think we can do that. Something like `cmake -DPYTHON_VERSION=3.9` with 3.6 as default.

SweetVishnya:
Can u make it?)

JonathanSalwan:
> Can u make it?)

ok, i will.

SweetVishnya:
Thanks)

PR #1003, opened 2021-03-06
Performance improvements for concrete memory

ek0:
Hello everyone,

Here is a set of small improvements I've put together for better overall performances. I haven't touched to the symbolic memory manager yet as the main bottleneck I encountered was with concrete memory.

First, I've removed `std::unordered_map` and replaced it by [tsl::robin_map](https://github.com/Tessil/robin-map). It's been proven more efficient and uses std::vector as buckets which is more cache friendly.

Also, `robin_map` uses a power of two mask instead of a modulo. It is not only more performant, but also since our keys are addresses, we have high entropy on the lower bits we can use an identity hash function. This saves us 60-70% of compute time on lookup.

I have modified the original implementation of the `robin_map::reserve` function. This way we can pre-reserve on insertion. This help when inserting big byte arrays by only doing one rehash. Instead of triggering multiple allocation and rehash as we insert a new byte, we will only do it once at the beginning. Originally `robin_map` forced a rehash on every `reserve()` call. Now it follows the `unordered_map::reserve` behavior and only reserve if needed.

Second, I've refactored the `BitVector` API to improve code locality. I noticed we didn't really need to pass copy of `std::pair` as arguments to `setPair`. This improved the overall front-end latency by 10%.

Here are the numbers before and after. The target executable is a 200MB file with around 10 sections of various size.
image

And after:

image

As you can see, we only had 12% of instructions retiring, we are now at 57%. :)

For more information about these numbers you can check [this link](https://software.intel.com/content/www/us/en/develop/documentation/vtune-cookbook/top/methodologies/top-down-microarchitecture-analysis-method.html).

PS: I also had to modify the `testers/CMakeLists.txt` file. It throwing me an error on cmake 3.18.2, on windows. This seemed to fix the issue.

Thanks!


src/libtriton/includes/triton/hash.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include <triton/tritonTypes.hpp>
+
+namespace triton
+{
+
+template<typename T>
+class IdentityHash
+{
+public:
+  // KEy is unique and are be returned as is
```
SweetVishnya, line 13: We need proper docstrings here.

SweetVishnya (changes requested):
Great work! The code looks fine to me. Could you please point out the parts that you modified in robin map?

Also, let me check that nothing is broken on Sydr benchmarks. I can run them next week.

SweetVishnya:
Btw, CI tests seem to crash.

SweetVishnya:
Also, have you tried setting concrete values on demand via get memory callback?

ek0:
I noticed the CI tests crashing, not sure why yet. Seems like some python allocation issues on linux?

For the part i modified in robin_map it's these 2 lines: https://github.com/ek0/Triton/blob/99d9fe26da5788ac1090c308a985a63c93b70f45/src/libtriton/includes/tsl/robin_hash.h#L1086

As for concrete memory values on demand you mean as a workaround? In my case, since one of the section is pure VM bytecode that will not change it would be much easier for me to initialize it all at once. And more efficient since most of it end up accessed.

src/libtriton/includes/triton/hash.hpp
SweetVishnya: Resolved

ek0:
Hi again,

I've quickly run the python tests on Windows and the tests are passing:

```
PS D:\codes\Triton\build> cmake --build . --config Release --target test-python
Microsoft (R) Build Engine version 16.9.0-preview-20570-04+b04541221 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

  "No syscall32 to extract on your plateform"
  "No syscall64 to extract on your plateform"
  triton.vcxproj -> D:\codes\Triton\build\src\libtriton\Release\triton.dll
  .....................................................................................................................................................................................................................................................................................
  ...........................................................................................s.s...................................................................
  ----------------------------------------------------------------------
  Ran 438 tests in 344.062s

  OK (skipped=2)
```

Could the failures on the CI be because we're hitting the memory limits of the CI machines (thus the `std::bad_alloc`)? If you were close to the limit before, this could be happen when a rehash is performed. `tsl::robin_map` will allocate a new_table to fit the actual content and move the old table to the new one. This process is a bit faster but can temporarily use a bit more memory than `std::unordered_map`.

src/libtriton/arch/x86/x8664Cpu.cpp
```diff
@@ -15,6 +15,7 @@
 #include <triton/immediate.hpp>
 #include <triton/x8664Cpu.hpp>
 
+#include <iterator>
```
JonathanSalwan: Why this include?

src/libtriton/includes/triton/aarch64Cpu.hpp
```diff
@@ -24,7 +24,8 @@
 #include <triton/register.hpp>
 #include <triton/tritonTypes.hpp>
 
-
+#include <triton/hash.hpp>
+#include <tsl/robin_map.h>
```
JonathanSalwan: What about defining `tsl::robin_map` into `triton::extlibs` (see [triton/externalLibs.hpp](https://github.com/JonathanSalwan/Triton/blob/master/src/libtriton/includes/triton/externalLibs.hpp))?. As we modified the `tsl::robin_map` behavior, we should keep it private for `triton` internals and let the user use the `tsl::robin_map` of his system for his C++ triton tools.

src/libtriton/includes/triton/hash.hpp
```diff
@@ -0,0 +1,19 @@
...
+  T operator()(const T& key) const { return key; }
```
JonathanSalwan: I think we can delete this file and define the `IdentityHash` into `tritonTypes.hpp`. Also, we may add to the comment something saying that this is used by `robin_map`.

src/libtriton/includes/triton/bitsVector.hpp
```diff
@@ -73,7 +73,7 @@ namespace triton {
         TRITON_EXPORT void setLow(triton::uint32 v);
 
         //! Sets the pair<high, low> position
-        TRITON_EXPORT void setPair(std::pair<triton::uint32, triton::uint32> p);
+        void setPair(triton::uint32 high, triton::uint32 low);
```
JonathanSalwan: I think we should find a new name now :D. What about:

* setBit(s)Position
* setSize
* setBit(s)
* ?

src/libtriton/arch/x86/x8664Cpu.cpp
ek0: I just forgot it, part of some tests. It should be removed in the last commit?

JonathanSalwan:
Regarding the CI crashs, it crashes on Windows and Linux.

> If you were close to the limit before

I'm not sure about that. It's probably due to a modification we did with this PR. Anyway, we have to get full green, i can't scarify my CI :D.

src/libtriton/includes/triton/bitsVector.hpp
SweetVishnya: I suggest setValue. It seems to be most close.

SweetVishnya: Modified the message.

SweetVishnya:
It should also become full green before I run Sydr benchmarks) Build Triton for Linux may help or even Windows subsystem for Linux.

ek0:
As discussed with @JonathanSalwan, I've added a compiler switch to let the user decide if he wants to use `tsl::robin_map` instead of `std::unordered_map`.
Since `std::unordered_map` uses linked list, lookup can be slower but rehashing is less expensive in terms of memory. While `tsl::robin_map` has been proven a lot more effective especially on big binaries, but rehashing can temporarily use a lot more memory. Overall, on my target example: `tsl::robin_map` will take 20s while `std::unordered_map` will take 27s to load.

As we are using `triton::byte` as value, the cost in memory already explodes (for every byte you roughly have to store at least 16+ bytes?) along with it. There's definitely something that can be improved on that side, but it'll go to another PR.

SweetVishnya:
We still need to make tests green with robin map on.

Issue #1006, opened 2021-03-12
Disassemble a set of instructions from concrete memory


JonathanSalwan:

JonathanSalwan:
```python
>>> from triton import *
>>> ctx = TritonContext(ARCH.X86_64)

>>> ctx.setConcreteMemoryAreaValue(0x1000, b"\x48\xb9\x88\x77\x66\x55\x44\x33\x22\x11\x48\xff\xc1\x48\x89\xc8\xc9\xc3")

>>> # Disassemble from memory
>>> insts = ctx.disassembly(0x1000, 5)
>>> for inst in insts:
...     print(inst)
... 
0x1000: movabs rcx, 0x1122334455667788
0x100a: inc rcx
0x100d: mov rax, rcx
0x1010: leave
0x1011: ret

# Disassembled instructions can be processed 
>>> for inst in insts:
...     ctx.processing(inst)
... 
True
True
True
True
True

>>> ctx.getSymbolicRegister(ctx.registers.rax)
(define-fun ref!9 () (_ BitVec 64) ref!2) ; MOV operation - 0x100d: mov rax, rcx
```

SweetVishnya:
Would this api be able to take pc value from concrete memory to handle next address after jump?

JonathanSalwan:
> Would this api be able to take pc value from concrete memory to handle next address after jump?

I think this is two kinds of different disassembly. Btw, to take in account the pc value, the instruction must be executed.

SweetVishnya:
This api can stop on branch instruction if instruction count is not provided.

PR #1007, opened 2021-03-12
Disassemble to control flow instruction from concrete memory

SweetVishnya:
Hi!

I implemented API to disassemble multiple instructions and stop on a control flow instruction. This may be useful to execute code by blocks.

xref #1006

### DIFF
--- a/src/libtriton/arch/architecture.cpp
+++ b/src/libtriton/arch/architecture.cpp
@@ -233,16 +233,45 @@ namespace triton {
     }
 
 
-    std::list<triton::arch::Instruction> Architecture::disassembly(triton::uint64 addr, triton::usize count) const {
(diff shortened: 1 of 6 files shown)


src/libtriton/arch/architecture.cpp
```diff
@@ -233,13 +233,20 @@ namespace triton {
     }
 
 
-    std::list<triton::arch::Instruction> Architecture::disassembly(triton::uint64 addr, triton::usize count) const {
-      std::list<triton::arch::Instruction> ret;
+    std::vector<triton::arch::Instruction> Architecture::disassembly(triton::uint64 addr, triton::usize count) const {
+      std::vector<triton::arch::Instruction> ret;
```
JonathanSalwan, line 238: `ret.reserve(count)`?

PR #1033, opened 2021-07-26
AST optimization for ite node: True ? A : B => A, False ? A : B => B

SweetVishnya:
Simplify constant ite condition.

### DIFF
--- a/src/libtriton/ast/astContext.cpp
+++ b/src/libtriton/ast/astContext.cpp
@@ -842,6 +842,13 @@ namespace triton {
 
 
     SharedAbstractNode AstContext::ite(const SharedAbstractNode& ifExpr, const SharedAbstractNode& thenExpr, const SharedAbstractNode& elseExpr) {
+      if (this->modes->isModeEnabled(triton::modes::AST_OPTIMIZATIONS)) {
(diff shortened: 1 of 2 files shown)


JonathanSalwan:
I think it's more a `CONSTANT_FOLDING` mode than an `AST_OPTIMIZATION`? We should maybe refactor [these lines](https://github.com/JonathanSalwan/Triton/blob/master/src/libtriton/ast/astContext.cpp#L850-L854)? (i do not remember why i did this: `node->isLogical() == false`)

SweetVishnya:
Can I enable it for both modes cause we do not use `CONSTANT_FOLDING`?

Maybe we can just remove [those lines](https://github.com/JonathanSalwan/Triton/blob/master/src/libtriton/ast/astContext.cpp#L850-L854)? Because constants for then and else should be already folded by the time.

RobinDavid:
Thanks @SweetVishnya . Yes it would be nice to have it in both mode. The difference between AST_OPTIMIZATION and CONSTANT_FOLDING is rather tight..

SweetVishnya:
Fixed review issues in new commit

PR #1034, opened 2021-08-05
use vcpkg to reference libraries

ZhouZiY:


JonathanSalwan:
Thanks for this MR. However, there is issue with builds on Linux and OSX. Could you fix that?

ZhouZiY:
I try to fix it. Now it ref z3 and capstone from vcpkg if  CMAKE_TOOLCHAIN_FILE defined.

Issue #1035, opened 2021-08-12
Crash when calling callback that is a nested python function


jordan9001:
```python
from triton import *

def setup_callback(ctx):
    def getMem(triton_ctx, memacc):
        print("In callback")
    ctx.addCallback(CALLBACK.GET_CONCRETE_MEMORY_VALUE, getMem)

ctx = TritonContext(ARCH.X86_64)
setup_callback(ctx)
ctx.getConcreteMemoryAreaValue(0x1000, 0x1)
```

This will crash with a NULL pointer dereference in `_PyFunction_Vectorcall` from `triton::callbacks::Callbacks::processCallbacks`.

A workaround for now is to simply not have callbacks that are nested functions.

JonathanSalwan:
Thx for reporting this! Do you know how to detect inner functions with CPython?

jordan9001:
I don't. But if I end up looking into it I will let you know!

SweetVishnya:
Doesn't it break the nested function lifetime? Is nested function really valid after returning from its parent function?

jordan9001:
This is fine python, afaik. This is a common way to implement a closure.

JonathanSalwan:
> This is fine python, afaik. This is a common way to implement a closure.

Mmmh, i think i agree with @SweetVishnya. For me the lifetime of the `getMem` function is over after returning from `setup_callback`. So when you record the callback you save a ptr which will be no longer valid after the return of `setup_callback`?

SweetVishnya:
Or, we should somehow increment and decrement function object ref count inside Python API...

jordan9001:
I think you are right @SweetVishnya. The crash no longer happens when I manually save a reference in python:
```python
from triton import *

def setup_callback(ctx):
    def getMem(triton_ctx, memacc):
        print("In callback")
    ctx.addCallback(CALLBACK.GET_CONCRETE_MEMORY_VALUE, getMem)
    return getMem

ctx = TritonContext(ARCH.X86_64)
ref = setup_callback(ctx)
ctx.getConcreteMemoryAreaValue(0x1000, 0x1)
```

jordan9001:
I was able to fix this with the following patch:
```diff
diff --git a/src/libtriton/bindings/python/objects/pyTritonContext.cpp b/src/libtriton/bindings/python/objects/pyTritonContext.cpp
index 3a9f0527..30221e0d 100644
[...]

PR #1038, opened 2021-08-15
added a fix for reference counting with python callbacks

jordan9001:
PR created in response to https://github.com/JonathanSalwan/Triton/issues/1035

This has more changes than the patch referenced in that issue. This causes Triton to now throw an exception if you try to remove a callback and do not find the callback. If the same callback is registered multiple times, previously one call to remove would remove all the duplicates as well. With this update remove must be called the same number of times as add.

This was added because it is necessary to prevent bad dereferences on python objects passed to `removeCallback`.

### DIFF
--- a/src/libtriton/arch/x86/x86Specifications.cpp
+++ b/src/libtriton/arch/x86/x86Specifications.cpp
@@ -50,7 +50,7 @@ namespace triton {
                                                       X86_LOWER,                                            \
                                                       true)                                                 \
                               );                                                                            \
-            name2id.emplace(#LOWER_NAME, ID_REG_X86_##UPPER_NAME);
+          name2id.emplace(#LOWER_NAME, ID_REG_X86_##UPPER_NAME);
           // Handle register not available in capstone as normal registers
           #define REG_SPEC_NO_CAPSTONE REG_SPEC
           #include "triton/x86.spec"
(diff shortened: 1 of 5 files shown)


src/libtriton/arch/x86/x86Specifications.cpp
```diff
@@ -50,7 +50,7 @@ namespace triton {
                                                       X86_LOWER,                                            \
                                                       true)                                                 \
                               );                                                                            \
-            name2id.emplace(#LOWER_NAME, ID_REG_X86_##UPPER_NAME);
+          name2id.emplace(#LOWER_NAME, ID_REG_X86_##UPPER_NAME);
```
jordan9001, line 53: This was done to remove compile time warnings. This doesn't change the way the code works, but if this _was_ intended to be part of the above if statement, then brackets need to be added.

JonathanSalwan:
Thanks for this MR.

Looks like there is an unittest that does not pass (src/testers/unittests/test_callback.py). This unittest tests if there is a memory leak by counting Python references. Maybe we added to much `Py_INCREF` or miss one `Py_DECCREF`.

```python
def method_callback(self, api, _):                                          
     pass

def test_method_callback_removal(self):                                     
    # regression test for #809                                              
    import sys                                                              
    self.Triton = TritonContext()                                           
    self.Triton.setArchitecture(ARCH.X86_64)                                
    cb = self.method_callback.__func__                                      
    cb_initial_refcnt = sys.getrefcount(cb)                                 
    self.Triton.addCallback(CALLBACK.GET_CONCRETE_MEMORY_VALUE, self.method_callback)
    self.Triton.removeCallback(CALLBACK.GET_CONCRETE_MEMORY_VALUE, self.method_callback)
    self.assertTrue(cb_initial_refcnt == sys.getrefcount(cb))               
```

jordan9001:
With the updates I think it is correct. is there anything else you need from me @JonathanSalwan ?

JonathanSalwan:
Looks ok for me. Thanks @jordan9001! I will merge it this week :)

Cheers,

PR #1044, opened 2021-09-08
Support overriding of global solving timeout per request

SweetVishnya:
The motivation of this PR is following. We would like to solve queries in parallel. And sometimes we need to change solving timeout for specific queries (e.g, to detect symbolic address bounds with a small timeout). Global timeout does not allow to have different timeout for parallel queries.

### DIFF
--- a/src/libtriton/api/api.cpp
+++ b/src/libtriton/api/api.cpp
@@ -1089,21 +1089,21 @@ namespace triton {
   }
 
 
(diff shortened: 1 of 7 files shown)


JonathanSalwan:
Thanks for this MR. Looks good to me. Just one thing: can you add a comment to the timeout attribute of the `Z3Solver` class explaining what you said about global timeout and that some queries may need personal timeout.

```cpp
      /*! \brief Solver engine using z3. */
      class Z3Solver : public SolverInterface {
        private:
          //! The SMT solver timeout. By default, unlimited.
          triton::uint32 timeout;
```

Cheers!

SweetVishnya:
Added docs.

PR #1046, opened 2021-09-14
Modernize fetching the python install directory

jamesjer:
The current way of fetching the python install directory gets the wrong one on systems where `get_python_lib()` (the general directory) differs from `get_python_lib(True)` (the platform-specific directory).  For example, on 64-bit Fedora Linux systems, the former is `/usr/lib/python[version]/site-packages` while the latter is `/usr/lib64/python[version]/site-packages`.

This commit:
- removes `from __future__ import print_function`, which is only needed for python 2
- uses `sysconfig.get_path` instead of `distutils.sysconfig.get_python_lib`.  The former has been available since python 3.2; the latter has been deprecated and will be removed in some future version of python.

### DIFF
--- a/src/libtriton/CMakeLists.txt
+++ b/src/libtriton/CMakeLists.txt
@@ -360,7 +360,7 @@ if(PYTHON_BINDINGS)
         DEPENDS triton
     )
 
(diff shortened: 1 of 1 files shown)


JonathanSalwan:
Thanks for this MR ! :)

PR #1048, opened 2021-10-02
Update macOS installation doc

abcSup:
Update the required libpython version and cmake command for macOS in doc


JonathanSalwan:
Thanks :)

PR #1060, opened 2021-12-10
Instruction semantics

apach301:
- Add semantics for some x86 instructions
- Improve check_semantics.py to check values of destination memory operands
- Fix semantics for xadd instruction

### DIFF
--- a/src/libtriton/arch/x86/x86Semantics.cpp
+++ b/src/libtriton/arch/x86/x86Semantics.cpp
@@ -370,6 +370,7 @@ VPMULHW                      | avx/avx2   | VEX Multiply Packed Signed Integers 
 VPMULLW                      | avx/avx2   | VEX Multiply Packed Signed Integers and Store Low Result
 VPOR                         | avx/avx2   | VEX Logical OR
 VPSHUFD                      | avx/avx2   | VEX Shuffle Packed Doublewords
+VPSIGNW                      | avx/avx2   | VEX Packed SIGN
 VPSLLDQ                      | avx/avx2   | VEX Shift Packed Double Quadword Left Logical
 VPSLLW                       | avx/avx2   | VEX Shift Packed Word Left Logical
 VPSRAD                       | avx/avx2   | VEX Shift Packed Doubleword Right Arithmetic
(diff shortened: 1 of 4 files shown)


JonathanSalwan:
Wow! Thanks for this MR! Let me few days to review all the code :)

src/libtriton/arch/x86/x86Semantics.cpp
```diff
@@ -14696,6 +14698,70 @@ namespace triton {
       }
 
 
+      void x86Semantics::vperm2i128_s(triton::arch::Instruction& inst) {
```
apach301, line 14830: Instruction `vperm2i128 dst, src1, src2, src3` - an avx2 [instruction](https://www.laruence.com/x86/VPERM2I128.html) that permutes xmm-parts of src1 and src2 registers (both ymm) into dst. Permutation is defined by control byte src3 (imm8). Destination can be completely zeroed or consisted only from one of the srcs according to src3. Hence non-trivial taint handling used here.

src/libtriton/arch/x86/x86Semantics.cpp
```diff
@@ -9913,6 +9919,132 @@ namespace triton {
...
+      void x86Semantics::pmulhw_s(triton::arch::Instruction& inst) {
+        auto& src1 = inst.operands[0];
+        auto& src2 = inst.operands[1];
+
+        /* Create symbolic operands */
+        auto op1 = this->symbolicEngine->getOperandAst(inst, src1);
+        auto op2 = this->symbolicEngine->getOperandAst(inst, src2);
+
+        /* Create the semantics */
+        if ((!op1->isSymbolized() && op1->evaluate() == 0) ||
```
apach301, line 10123: Handling case when the result is concretely zeroed. Same in `pmulld_s`, `pmullw_s` and `vpmullw_s`, `vpmulhw_s`.

src/testers/pin/check_semantics.py
```diff
@@ -119,6 +120,18 @@ def cafter(instruction):
                 print("     Native address   : %016x" %(nativeAddress))
                 print("     Symbolic address : %016x" %(astAddress))
 
+    # Check destination memory accesses
+    for mem, sym_mem in instruction.getStoreAccess():
```
apach301, line 124: Add check for destination memory access - to be able to verify that instruction writes to the memory correct values.

src/libtriton/arch/x86/x86Semantics.cpp
```diff
@@ -16563,44 +16563,29 @@ namespace triton {
...
-        expr1->isTainted = this->taintEngine->setTaint(dst, srcT);
-        expr2->isTainted = this->taintEngine->setTaint(src, dstT);
-
-        /* Create symbolic operands */
-        op1 = this->symbolicEngine->getOperandAst(inst, dst);
-        op2 = this->symbolicEngine->getOperandAst(inst, src);
-
-        /* Create the semantics */
-        auto node3 = this->astCtxt->bvadd(op1, op2);
+        auto node = this->astCtxt->bvadd(op1, op2);
```
apach301, line 16572: The `xadd` instruction started to fail on tests after implementation of destination memory operand check. The semantics was correct but somehow the symbolic expressions for the dst operand (in case of memory operand) was wrong. So I decided to rewrite semantics without dividing its logic on two steps (xchg and add) and calling `createSymbolicExpression()` only when final semantics is built - that fixes the tests.

PR #1062, opened 2021-12-22
Add Bitwuzla solver

apach301:
Here is some preliminary support of [Bitwuzla](https://github.com/bitwuzla/bitwuzla) solver.
TODO list:

- [x] Support Bitwuzla in build system
- [x] Implement Python bindings
- [ ] Add some tests

### DIFF
--- a/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
+++ b/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
@@ -1,0 +1,248 @@
+//! \file
+/*
+**  Copyright (C) - Triton
+**
+**  This program is under the terms of the Apache License 2.0.
+*/
+
+#include <vector>
+
+#include <triton/cpuSize.hpp>
+#include <triton/exceptions.hpp>
+#include <triton/symbolicExpression.hpp>
+#include <triton/symbolicVariable.hpp>
+#include <triton/tritonToBitwuzlaAst.hpp>
+
+
+
+namespace triton {
+  namespace ast {
+
+    TritonToBitwuzlaAst::TritonToBitwuzlaAst() {}
+
+    TritonToBitwuzlaAst::~TritonToBitwuzlaAst() {
+      this->translated_nodes.clear();
+      this->variables.clear();
+      this->symbols.clear();
+    }
+
+    const BitwuzlaTerm* TritonToBitwuzlaAst::convert(const SharedAbstractNode& node, Bitwuzla* bzla) {
+      auto nodes = childrenExtraction(node, true /* unroll*/, true /* revert */);
+      for (auto&& n : nodes) {
+        translated_nodes[n] = translate(n, bzla);
+      }
+      return translated_nodes.at(node);
+    }
+
+    const BitwuzlaTerm* TritonToBitwuzlaAst::translate(const SharedAbstractNode& node, Bitwuzla* bzla) {
+      if (node == nullptr)
+        throw triton::exceptions::AstTranslations("TritonToBitwuzlaAst::translate(): node cannot be null.");
+
+      std::vector<const BitwuzlaTerm*> children;
+      for (auto&& n : node->getChildren()) {
+        children.emplace_back(translated_nodes.at(n));
+      }
+
+      switch (node->getType()) {
+        case BVADD_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_ADD, children[0], children[1]);
+
+        case BVAND_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_AND, children[0], children[1]);
+
+        case BVASHR_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_ASHR, children[0], children[1]);
+
+        case BVLSHR_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_SHR, children[0], children[1]);
+
+        case BVMUL_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_MUL, children[0], children[1]);
+
+        case BVNAND_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_NAND, children[0], children[1]);
+
+        case BVNEG_NODE:
+          return bitwuzla_mk_term1(bzla, BITWUZLA_KIND_BV_NEG, children[0]);
+
+        case BVNOR_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_NOR, children[0], children[1]);
+
+        case BVNOT_NODE:
+          return bitwuzla_mk_term1(bzla, BITWUZLA_KIND_BV_NOT, children[0]);
+
+        case BVOR_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_OR, children[0], children[1]);
+
+        case BVROL_NODE: {
+          auto childNodes = node->getChildren();
+          auto idx = reinterpret_cast<IntegerNode*>(childNodes[1].get())->getInteger().convert_to<size_t>();
+          return bitwuzla_mk_term1_indexed1(bzla, BITWUZLA_KIND_BV_ROLI, children[0], idx);
+        }
+
+        case BVROR_NODE: {
+          auto childNodes = node->getChildren();
+          auto idx = reinterpret_cast<IntegerNode*>(childNodes[1].get())->getInteger().convert_to<size_t>();
+          return bitwuzla_mk_term1_indexed1(bzla, BITWUZLA_KIND_BV_RORI, children[0], idx);
+        }
+
+        case BVSDIV_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_SDIV, children[0], children[1]);
+
+        case BVSGE_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_SGE, children[0], children[1]);
+
+        case BVSGT_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_SGT, children[0], children[1]);
+
+        case BVSHL_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_SHL, children[0], children[1]);
+
+        case BVSLE_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_SLE, children[0], children[1]);
+
+        case BVSLT_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_SLT, children[0], children[1]);
+
+        case BVSMOD_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_SMOD, children[0], children[1]);
+
+        case BVSREM_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_SREM, children[0], children[1]);
+
+        case BVSUB_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_SUB, children[0], children[1]);
+
+        case BVUDIV_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_UDIV, children[0], children[1]);
+
+        case BVUGE_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_UGE, children[0], children[1]);
+
+        case BVUGT_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_UGT, children[0], children[1]);
+
+        case BVULE_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_ULE, children[0], children[1]);
+
+        case BVULT_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_ULT, children[0], children[1]);
+
+        case BVUREM_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_UREM, children[0], children[1]);
+
+        case BVXNOR_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_XNOR, children[0], children[1]);
+
+        case BVXOR_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_BV_XOR, children[0], children[1]);
+
+        case BV_NODE: {
+          auto childNodes = node->getChildren();
+          auto bv_size = reinterpret_cast<IntegerNode*>(childNodes[1].get())->getInteger().convert_to<size_t>();
+          auto sort = bv_sorts.find(bv_size);
+          if (sort == bv_sorts.end()) {
+            sort = bv_sorts.insert({bv_size, bitwuzla_mk_bv_sort(bzla, bv_size)}).first;
+          }
+
+          // Handle bitvector value as integer if it small enough.
+          if (bv_size <= sizeof(uint64_t) * 8) {
+            auto bv_value = reinterpret_cast<IntegerNode*>(childNodes[0].get())->getInteger().convert_to<uint64_t>();
+            return bitwuzla_mk_bv_value_uint64(bzla, sort->second, bv_value);
+          }
+          auto bv_value = reinterpret_cast<IntegerNode*>(childNodes[0].get())->getInteger().convert_to<std::string>();
+          return bitwuzla_mk_bv_value(bzla, sort->second, bv_value.c_str(), BITWUZLA_BV_BASE_DEC);
+        }
+
+        case CONCAT_NODE:
+          return bitwuzla_mk_term(bzla, BITWUZLA_KIND_BV_CONCAT, children.size(), children.data());
+
+        case DISTINCT_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_DISTINCT, children[0], children[1]);
+
+        case EQUAL_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_EQUAL, children[0], children[1]);
+
+        case EXTRACT_NODE: {
+          auto childNodes = node->getChildren();
+          auto high = reinterpret_cast<IntegerNode*>(childNodes[0].get())->getInteger().convert_to<size_t>();
+          auto low = reinterpret_cast<IntegerNode*>(childNodes[1].get())->getInteger().convert_to<size_t>();
+          return bitwuzla_mk_term1_indexed2(bzla, BITWUZLA_KIND_BV_EXTRACT, children[2], high, low);
+        }
+
+        case FORALL_NODE:
+          throw triton::exceptions::AstTranslations("TritonToBitwuzlaAst::translate(): FORALL node can't be converted due to a Bitwuzla issue (see #1062).");
+
+        case IFF_NODE:
+          return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_IFF, children[0], children[1]);
+
+        case INTEGER_NODE:
+          return nullptr;
+
+        case ITE_NODE:
+          return bitwuzla_mk_term3(bzla, BITWUZLA_KIND_ITE, children[0], children[1], children[2]);
+
+        case LAND_NODE:
+          return bitwuzla_mk_term(bzla, BITWUZLA_KIND_AND, children.size(), children.data());
+
+        case LET_NODE: {
+          auto childNodes = node->getChildren();
+          symbols[reinterpret_cast<triton::ast::StringNode*>(childNodes[0].get())->getString()] = childNodes[1];
+          return children[2];
+        }
+
+        case LNOT_NODE:
+          return bitwuzla_mk_term1(bzla, BITWUZLA_KIND_NOT, children[0]);
+
+        case LOR_NODE:
+          return bitwuzla_mk_term(bzla, BITWUZLA_KIND_OR, children.size(), children.data());
+
+        case LXOR_NODE:
+          return bitwuzla_mk_term(bzla, BITWUZLA_KIND_XOR, children.size(), children.data());
+
+        case REFERENCE_NODE: {
+          auto ref = reinterpret_cast<ReferenceNode*>(node.get())->getSymbolicExpression()->getAst();
+          return translated_nodes.at(ref);
+        }
+
+        case STRING_NODE: {
+          std::string value = reinterpret_cast<triton::ast::StringNode*>(node.get())->getString();
+
+          auto it = symbols.find(value);
+          if (it == symbols.end())
+            throw triton::exceptions::AstTranslations("TritonToBitwuzlaAst::translate(): [STRING_NODE] Symbols not found.");
+
+          return translated_nodes.at(it->second);
+        }
+
+        case SX_NODE: {
+          auto childNodes = node->getChildren();
+          auto ext = reinterpret_cast<IntegerNode*>(childNodes[0].get())->getInteger().convert_to<size_t>();
+          return bitwuzla_mk_term1_indexed1(bzla, BITWUZLA_KIND_BV_SIGN_EXTEND, children[1], ext);
+        }
+
+        case VARIABLE_NODE: {
+          const auto& symVar = reinterpret_cast<VariableNode*>(node.get())->getSymbolicVariable();
+          auto size = symVar->getSize();
+          auto sort = bv_sorts.find(size);
+          if (sort == bv_sorts.end()) {
+            sort = bv_sorts.insert({size, bitwuzla_mk_bv_sort(bzla, size)}).first;
+          }
+          auto n = bitwuzla_mk_const(bzla, sort->second, symVar->getName().c_str());
+          variables[n] = symVar;
+          return n;
+        }
+
+        case ZX_NODE: {
+          auto childNodes = node->getChildren();
+          auto ext = reinterpret_cast<IntegerNode*>(childNodes[0].get())->getInteger().convert_to<size_t>();
+          return bitwuzla_mk_term1_indexed1(bzla, BITWUZLA_KIND_BV_ZERO_EXTEND,children[1], ext);
+        }
+
+        default:
+          throw triton::exceptions::AstTranslations("TritonToBitwuzlaAst::translate(): Invalid kind of node.");
+      }
+    }
+
+  }; /* ast namespace */
+}; /* triton namespace */
--- a/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
+++ b/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
@@ -21,17 +21,25 @@ namespace triton {
     TritonToBitwuzlaAst::TritonToBitwuzlaAst() {}
 
     TritonToBitwuzlaAst::~TritonToBitwuzlaAst() {
-      this->translated_nodes.clear();
+      this->translatedNodes.clear();
       this->variables.clear();
       this->symbols.clear();
+    }
+
+    const std::unordered_map<const BitwuzlaTerm*, triton::engines::symbolic::SharedSymbolicVariable>& TritonToBitwuzlaAst::getVariables() const {
+      return this->variables;
+    }
+
+    const std::map<size_t, const BitwuzlaSort*>& TritonToBitwuzlaAst::getBitvectorSorts() const {
+      return this->bvSorts;
     }
 
     const BitwuzlaTerm* TritonToBitwuzlaAst::convert(const SharedAbstractNode& node, Bitwuzla* bzla) {
       auto nodes = childrenExtraction(node, true /* unroll*/, true /* revert */);
       for (auto&& n : nodes) {
-        translated_nodes[n] = translate(n, bzla);
+        translatedNodes[n] = translate(n, bzla);
       }
-      return translated_nodes.at(node);
+      return translatedNodes.at(node);
     }
 
     const BitwuzlaTerm* TritonToBitwuzlaAst::translate(const SharedAbstractNode& node, Bitwuzla* bzla) {
@@ -40,7 +48,7 @@ namespace triton {
 
       std::vector<const BitwuzlaTerm*> children;
       for (auto&& n : node->getChildren()) {
-        children.emplace_back(translated_nodes.at(n));
+        children.emplace_back(translatedNodes.at(n));
       }
 
       switch (node->getType()) {
@@ -140,9 +148,9 @@ namespace triton {
         case BV_NODE: {
           auto childNodes = node->getChildren();
           auto bv_size = reinterpret_cast<IntegerNode*>(childNodes[1].get())->getInteger().convert_to<size_t>();
-          auto sort = bv_sorts.find(bv_size);
-          if (sort == bv_sorts.end()) {
-            sort = bv_sorts.insert({bv_size, bitwuzla_mk_bv_sort(bzla, bv_size)}).first;
+          auto sort = bvSorts.find(bv_size);
+          if (sort == bvSorts.end()) {
+            sort = bvSorts.insert({bv_size, bitwuzla_mk_bv_sort(bzla, bv_size)}).first;
           }
 
           // Handle bitvector value as integer if it small enough.
@@ -202,7 +210,7 @@ namespace triton {
 
         case REFERENCE_NODE: {
           auto ref = reinterpret_cast<ReferenceNode*>(node.get())->getSymbolicExpression()->getAst();
-          return translated_nodes.at(ref);
+          return translatedNodes.at(ref);
         }
 
         case STRING_NODE: {
@@ -212,7 +220,7 @@ namespace triton {
           if (it == symbols.end())
             throw triton::exceptions::AstTranslations("TritonToBitwuzlaAst::translate(): [STRING_NODE] Symbols not found.");
 
-          return translated_nodes.at(it->second);
+          return translatedNodes.at(it->second);
         }
 
         case SX_NODE: {
@@ -224,9 +232,9 @@ namespace triton {
         case VARIABLE_NODE: {
           const auto& symVar = reinterpret_cast<VariableNode*>(node.get())->getSymbolicVariable();
           auto size = symVar->getSize();
-          auto sort = bv_sorts.find(size);
-          if (sort == bv_sorts.end()) {
-            sort = bv_sorts.insert({size, bitwuzla_mk_bv_sort(bzla, size)}).first;
+          auto sort = bvSorts.find(size);
+          if (sort == bvSorts.end()) {
+            sort = bvSorts.insert({size, bitwuzla_mk_bv_sort(bzla, size)}).first;
           }
           auto n = bitwuzla_mk_const(bzla, sort->second, symVar->getName().c_str());
           variables[n] = symVar;
--- a/src/libtriton/bindings/python/namespaces/initSolverNamespace.cpp
+++ b/src/libtriton/bindings/python/namespaces/initSolverNamespace.cpp
@@ -38,7 +38,12 @@ namespace triton {
       void initSolverNamespace(PyObject* solverDict) {
         PyDict_Clear(solverDict);
 
+        #if defined(TRITON_Z3_INTERFACE)
         xPyDict_SetItemString(solverDict, "Z3", PyLong_FromUint32(triton::engines::solver::SOLVER_Z3));
+        #endif
+        #if defined(TRITON_BITWUZLA_INTERFACE)
+        xPyDict_SetItemString(solverDict, "BITWUZLA", PyLong_FromUint32(triton::engines::solver::SOLVER_BITWUZLA));
+        #endif
       }
 
     }; /* python namespace */
--- a/src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
+++ b/src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
@@ -1,0 +1,245 @@
+//! \file
+/*
+**  Copyright (C) - Triton
+**
+**  This program is under the terms of the Apache License 2.0.
+*/
+
+#include <fstream>
+#include <regex>
+#include <string>
+
+#include <triton/astContext.hpp>
+#include <triton/bitwuzlaSolver.hpp>
+#include <triton/exceptions.hpp>
+#include <triton/solverModel.hpp>
+#include <triton/symbolicExpression.hpp>
+#include <triton/symbolicVariable.hpp>
+#include <triton/tritonToBitwuzlaAst.hpp>
+#include <triton/tritonTypes.hpp>
+
+
+namespace triton {
+  namespace engines {
+    namespace solver {
+
+      int32_t BitwuzlaSolver::terminate_fun(void* state) {
+        auto p = reinterpret_cast<SolverParams*>(state);
+
+        // Check time only once in every 1000 calls.
+        if (++p->call_cnt < 1000) {
+          return 0;
+        }
+        p->call_cnt = 0;
+
+        // Count delta.
+        auto delta = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now() - p->start).count();
+
+        // Check timeout expired.
+        if (p->timeout && delta > p->timeout) {
+          p->code = 1;
+          return 1;
+        }
+
+#if defined(__unix__)
+        // Check memory limit.
+        delta /= 1000;
+        if (p->memory_limit && delta > p->last_mem_check && delta % p->delay == 0) {
+          p->last_mem_check = delta;
+
+          size_t memory_usage = 0;
+          std::ifstream str("/proc/self/status");
+          if (str.is_open()) {
+            std::string line;
+            std::regex vmsize("VmRSS:\\s*([0-9]+)\\s*kB");
+            while (std::getline(str, line)) {
+              std::smatch match;
+              if (std::regex_match(line, match, vmsize)) {
+                memory_usage = strtoul(match.str(1).c_str(), NULL, 10) / 1024;
+                break;
+              }
+            }
+          }
+          if (!memory_usage) {
+            return 0;
+          }
+          if (memory_usage > p->memory_limit) {
+            p->code = 2;
+            return 2;
+          }
+          // The more memory is available the rarely we check current
+          // memory consumption in order to save resources.
+          if (memory_usage < p->memory_limit / 4) {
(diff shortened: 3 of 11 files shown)


cnheitman:
Nice!! I'll take a look and give it a try as soon as I can.

JonathanSalwan:
Awesome! I will take a look today. About `Support Bitwuzla in build system`, like z3, I think you can let the user builds his own dependencies.

JonathanSalwan (approved):

JonathanSalwan:
Just did the review, looks good at the first shot. I will install Bitwuzla and test the code.

cnheitman (approved):

cnheitman:
Same here, the code looks good. I have to test it a bit now.

cnheitman:
@apach301 What command line do you use to build it? I had to build it like this:

```bash
cmake -DBITWUZLA_INTERFACE=On -DBITWUZLA_INCLUDE_DIRS=/usr/local/include/bitwuzla \
      -DBITWUZLA_LIBRARIES=/usr/local/lib/libbitwuzla.so ..
```
Btw, I installed Bitwuzla system-wide, which correctly installed `BitwuzlaConfig.cmake`, but `cmake` doesn't seem to find it.

JonathanSalwan:
@cnheitman On my side, I compile Triton with the following command line:

```
cmake \
-DBITWUZLA_INTERFACE=On \
-DBITWUZLA_INCLUDE_DIRS=/usr/local/include/ \
-DBITWUZLA_LIBRARIES=/usr/local/lib/libbitwuzla.so \
..
```

@apach301 I got one compile issue:

```
[ 48%] Building CXX object src/libtriton/CMakeFiles/triton.dir/ast/bitwuzla/tritonToBitwuzlaAst.cpp.o
In file included from /home/jonathan/Works/Tools/Triton-bitwuzla-support/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp:14:
/home/jonathan/Works/Tools/Triton-bitwuzla-support/src/libtriton/includes/triton/tritonToBitwuzlaAst.hpp:68:14: error: ‘map’ in namespace ‘std’ does not name a template type
   68 |         std::map<size_t, const BitwuzlaSort*> bv_sorts;
      |              ^~~
In file included from /home/jonathan/Works/Tools/Triton-bitwuzla-support/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp:14:
/home/jonathan/Works/Tools/Triton-bitwuzla-support/src/libtriton/includes/triton/tritonToBitwuzlaAst.hpp:14:1: note: ‘std::map’ is defined in header ‘<map>’; did you forget to ‘#include <map>’?
   13 | #include <bitwuzla/bitwuzla.h>
  +++ |+#include <map>
   14 | 
/home/jonathan/Works/Tools/Triton-bitwuzla-support/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp: In member function ‘const BitwuzlaTerm* triton::ast::TritonToBitwuzlaAst::translate(const SharedAbstractNode&, Bitwuzla*)’:
/home/jonathan/Works/Tools/Triton-bitwuzla-support/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp:145:23: error: ‘bv_sorts’ was not declared in this scope
  145 |           auto sort = bv_sorts.find(size);
      |                       ^~~~~~~~
make[2]: *** [src/libtriton/CMakeFiles/triton.dir/build.make:688: src/libtriton/CMakeFiles/triton.dir/ast/bitwuzla/tritonToBitwuzlaAst.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1061: src/libtriton/CMakeFiles/triton.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

I've fixed the issue with the following patch:

```patch
diff --git a/src/libtriton/includes/triton/tritonToBitwuzlaAst.hpp b/src/libtriton/includes/triton/tritonToBitwuzlaAst.hpp
index e1eea95..ce0fb3d 100644
--- a/src/libtriton/includes/triton/tritonToBitwuzlaAst.hpp
+++ b/src/libtriton/includes/triton/tritonToBitwuzlaAst.hpp
@@ -8,6 +8,7 @@
 #ifndef TRITON_TRITONTOBITWIZLAAST_H
 #define TRITON_TRITONTOBITWIZLAAST_H
 
+#include <map>
 #include <unordered_map>
 
 #include <bitwuzla/bitwuzla.h>
```

JonathanSalwan:
To test your code I've hardcoded the Bitwuzla solver as default with the following patch:

```patch
diff --git a/src/libtriton/engines/solver/solverEngine.cpp b/src/libtriton/engines/solver/solverEngine.cpp
index 9c91a12..5f1103f 100644
--- a/src/libtriton/engines/solver/solverEngine.cpp
+++ b/src/libtriton/engines/solver/solverEngine.cpp
@@ -19,7 +19,7 @@ namespace triton {
         this->kind = triton::engines::solver::SOLVER_INVALID;
         #ifdef TRITON_Z3_INTERFACE
         /* By default we initialized the z3 solver */
-        this->setSolver(triton::engines::solver::SOLVER_Z3);
+        this->setSolver(triton::engines::solver::SOLVER_BITWUZLA);
         #endif
       }
```

Then I've executed some of CTF solving [examples](https://github.com/JonathanSalwan/Triton/tree/master/src/examples/python/ctf-writeups) and it looks working well :).

Then, i've executed all unit tests via `cmake check` and got this:

```
..............EEEE.E.EEEE.....E.EEEEEEEEEEEE.EEEEEEEEEE...................................................................................................................................................................F..F.....................[bitwuzla] bitwuzla_mk_term: expected variable term at index 0
```

cnheitman:
Hey @JonathanSalwan, in my case I solved the issue (`cmake` not finding `Bitwuzla`) doing this couple of tweaks:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index c56e0f56..07959ebf 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -189,7 +189,7 @@ if(BITWUZLA_INTERFACE)
         else()
             find_package(Bitwuzla REQUIRED)
         endif()
-        if(NOT BITWUZLA_FOUND)
+        if(NOT Bitwuzla_FOUND)
             message(FATAL_ERROR "Bitwuzla not found")
         endif()
     endif()
diff --git a/src/libtriton/CMakeLists.txt b/src/libtriton/CMakeLists.txt
index d10cdc0d..cc9cbe6a 100644
--- a/src/libtriton/CMakeLists.txt
+++ b/src/libtriton/CMakeLists.txt
@@ -354,7 +354,7 @@ target_link_libraries(
     ${PYTHON_LIBRARIES}
     ${Boost_LIBRARIES}
     ${Z3_LIBRARIES}
-    ${BITWUZLA_LIBRARIES}
+    Bitwuzla::bitwuzla
     ${CAPSTONE_LIBRARIES}
     ${LIBTRITON_OTHER_LIBS}
 )
```
This way you can compile it without having to specify the location (that is, `cmake -DBITWUZLA_INTERFACE=On ..`). Using this approach `BITWUZLA_LIBRARIES` and `BITWUZLA_INCLUDE_DIRS` are no longer needed (and in case you want to use a local installation of Bitwuzla you can specify it using the `-DBitwusla_DIR=<path>` command line option when invoking `cmake`).

BTW, I did not have any compile errors in my case.

Regarding the initialization of the solver, I ended up doing this:

```diff
diff --git a/src/libtriton/engines/solver/solverEngine.cpp b/src/libtriton/engines/solver/solverEngine.cpp
index 9c91a12d..8e38a9e6 100644
--- a/src/libtriton/engines/solver/solverEngine.cpp
+++ b/src/libtriton/engines/solver/solverEngine.cpp
@@ -17,9 +17,11 @@ namespace triton {
 
       SolverEngine::SolverEngine() {
         this->kind = triton::engines::solver::SOLVER_INVALID;
-        #ifdef TRITON_Z3_INTERFACE
+        #if defined(TRITON_Z3_INTERFACE)
         /* By default we initialized the z3 solver */
         this->setSolver(triton::engines::solver::SOLVER_Z3);
+        #elif defined(TRITON_BITWUZLA_INTERFACE)
+        this->setSolver(triton::engines::solver::SOLVER_BITWUZLA);
         #endif
       }
``` 
This is also a nice approach to instantiate Bitwusla in case you compile Triton with `-DZ3_INTERFACE=Off`.

JonathanSalwan:
@cnheitman nice about the cmake :). About the solver by default, yes you did right to instantiate bitwuzla if it's enabled and z3 not.

Still continuing to test the solver and got this:

```
$ python synthetizing_obfuscated_expressions.py
In: (((((SymVar_0 | SymVar_1) + SymVar_1) & 0xff) - ((~(SymVar_0) & 0xff) & SymVar_1)) & 0xff)
Out: ((SymVar_0 + SymVar_1) & 0xff)

In: (((((SymVar_0 | SymVar_1) - SymVar_1) & 0xff) + ((~(SymVar_0) & 0xff) & SymVar_1)) & 0xff)
Out: (SymVar_0 ^ SymVar_1)

In: ((SymVar_0 & (~(SymVar_1) & 0xff)) | ((~(SymVar_0) & 0xff) & SymVar_1))
Out: (SymVar_0 ^ SymVar_1)

In: (((((SymVar_0 ^ SymVar_1) + SymVar_1) & 0xff) - ((~(SymVar_0) & 0xff) & SymVar_1)) & 0xff)
Out: (SymVar_0 | SymVar_1)

In: (((((-((SymVar_0 | SymVar_1)) & 0xff) + SymVar_1) & 0xff) + SymVar_0) & 0xff)
Out: (SymVar_1 & SymVar_0)

In: (((((SymVar_2 << 0x8) & 0xffffffff) >> 0x10) << 0x8) & 0xffffffff)
[bitwuzla] bitwuzla_mk_term: expected variable term at index 0


$ python ./proving_opaque_predicates.py
xor eax, eax
jo 9
[1]    12760 segmentation fault  python ./proving_opaque_predicates.py
```

cnheitman:
Hey @JonathanSalwan , I was having the same issue on my side. I fixed the segfault one by adding the following check: 

```diff
diff --git a/src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp b/src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
index 9144e852..4b0133af 100644
--- a/src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
+++ b/src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
@@ -61,6 +61,7 @@ namespace triton {
         this->writeBackStatus(res, p.status, status);
 
         std::vector<std::unordered_map<triton::usize, SolverModel>> ret;
+        if (res == BITWUZLA_SAT) {
         auto symvar_size = bzlaAst->getVariables().begin()->second->getSize();
         auto symvar_sort = bitwuzla_mk_bv_sort(bzla, symvar_size);
 
@@ -93,6 +94,7 @@ namespace triton {
           // Decrement the limit.
           limit--;
         }
+        }
 
         bitwuzla_delete(bzla);
```

Regarding the other one, there seems to be an issue when it tries to build a `forall` node in `tritonToBitwuzlaAst.cpp`. I couldn't find anything odd at first sight. However, I'm not familiar with Bitwuzla's API, perhaps @apach301 can provide some insight with this one.

I'll continue doing some more tests on my side.

cnheitman:
One more update. I ended up skipping some Z3-specific tests so we can run current them successfully (until we add Bitwuzla-specific tests). I added the code below (quick and dirty hack for the moment) to `test_ast_conversion.py` (same for `test_ast_deep.py` and `test_ast_eval.py`):

```diff
diff --git a/src/testers/unittests/test_ast_conversion.py b/src/testers/unittests/test_ast_conversion.py
index cffbcd04..39963773 100644
--- a/src/testers/unittests/test_ast_conversion.py
+++ b/src/testers/unittests/test_ast_conversion.py
@@ -9,6 +9,12 @@ import utils
 
 from triton import *
 
+try:
+    Triton = TritonContext()
+    astCtxt = Triton.getAstContext()
+    Triton.evaluateAstViaZ3(astCtxt.bvtrue() == astCtxt.bvtrue())
+except:
+    raise unittest.SkipTest('Z3_INTERFACE=Off, skipping test.')
  

 class TestAstConversion(unittest.TestCase):
```

With these changes (the one above and the others I posted previously), only `synthetizing_obfuscated_expressions.py` is failing on my side (because of the exception that Bitwuzla throws). I think once we fixed that issue, we can update the settings of the CI to test Bitwuzla too (`-DZ3_INTERFACE=Off -DBITWUZLA_INTERFACE=On`), merge this PR to `dev-v0.9` and start working on the TODO list mentioned at the beginning of this issue. What do you think @JonathanSalwan ?

apach301:
I am not able to build Triton with only `-DBITWUZLA_INTERFACE=On`: it requires Bitwuzla’ include directory to compile. So I think it can be organized the same way as z3 - both via find_package (fixed up to `Bitwuzla`) and INCLUDE_DIRS/INCLUDE_LIBRARIES variables.

As for the unittests: with all proposed fixes only forall-tests are failing on my side. I tried to sort out the problem and this is what I found out: Bitwuzla solver have a poor support for the quantifiers right now. In more details, it turned out that bitwuzla’s `forall`-term requires a special variable-term definition for symbolic variable (I made all symbolic variables as constant-terms, because it is the only way bitwuzla can generate models). Even after I made a workaround for this problem (additional definition of variable-terms for each symbolic variables and following substitution of const-terms into variable-terms when forall-node is processed) bitwuzla refused to work because of the disabled quantifiers support. I tried manually patch bitwuzla code to process one quantified formula with forall-node - it returnes the correct model but it can be parsed via API the same way as other models (quantifiers don’t support certain API).

JonathanSalwan:
> One more update. I ended up skipping some Z3-specific tests so we can run current them successfully (until we add Bitwuzla-specific tests). I added the code below (quick and dirty hack for the moment) to `test_ast_conversion.py` (same for `test_ast_deep.py` and `test_ast_eval.py`):
> 
> ```diff
> diff --git a/src/testers/unittests/test_ast_conversion.py b/src/testers/unittests/test_ast_conversion.py
> index cffbcd04..39963773 100644
> --- a/src/testers/unittests/test_ast_conversion.py
> +++ b/src/testers/unittests/test_ast_conversion.py
> @@ -9,6 +9,12 @@ import utils
>  
>  from triton import *
>  
> +try:
> +    Triton = TritonContext()
> +    astCtxt = Triton.getAstContext()
> +    Triton.evaluateAstViaZ3(astCtxt.bvtrue() == astCtxt.bvtrue())
> +except:
> +    raise unittest.SkipTest('Z3_INTERFACE=Off, skipping test.')
>   
> 
>  class TestAstConversion(unittest.TestCase):
> ```

@cnheitman maybe we can rename `evaluateAstViaZ3` by `evaluateAstViaSolver` if Bitwuzla allow us to evaluate an AST to a concrete value if there is no sym var?

> With these changes (the one above and the others I posted previously), only `synthetizing_obfuscated_expressions.py` is failing on my side (because of the exception that Bitwuzla throws). I think once we fixed that issue, we can update the settings of the CI to test Bitwuzla too (`-DZ3_INTERFACE=Off -DBITWUZLA_INTERFACE=On`), merge this PR to `dev-v0.9` and start working on the TODO list mentioned at the beginning of this issue. What do you think @JonathanSalwan ?

Sounds a nice plan :)

JonathanSalwan:
> Thus, I have a question - is supporting forall-node is crucial for Triton or it can be disabled as simplify/evaliate functionality?

@apach301, I think if the support of quantifier is not reliable yet on Bitwuzla, we can simply ignore it and raise an exception like below:

```patch
diff --git a/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp b/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
index 2953cb4..5986196 100644
--- a/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
+++ b/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
@@ -166,7 +166,7 @@ namespace triton {
         }
 
         case FORALL_NODE:
-          return bitwuzla_mk_term(bzla, BITWUZLA_KIND_FORALL, children.size(), children.data());
+          throw triton::exceptions::AstTranslations("TritonToBitwuzlaAst::translate(): FORALL node can't be converted due to a Bitwuzla issue (see #1062).");
 
         case IFF_NODE:
           return bitwuzla_mk_term2(bzla, BITWUZLA_KIND_IFF, children[0], children[1]);
```

What do you think @cnheitman ?

SweetVishnya:
Let me please review the code today before merge. Also, there is one more TODO from me:

- [ ] Supporting memory limit (we can copy some code from Sydr that computes current process ram usage)

apach301:
Found out that bug mentioned in https://github.com/JonathanSalwan/Triton/pull/1062#issuecomment-1001586516 was in trying to process model for constant formula (without symbolic variables). Moreover, was discovered that solver's `check_sat()` function was called twice in `getModels()` with limit=1: https://github.com/JonathanSalwan/Triton/pull/1066#issue-811424594

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,182 @@
...
+      }
+
+
+      std::unordered_map<triton::usize, SolverModel> BitwuzlaSolver::getModel(const triton::ast::SharedAbstractNode& node, triton::engines::solver::status_e* status, triton::uint32 timeout) const {
+        std::unordered_map<triton::usize, SolverModel> ret;
+        std::vector<std::unordered_map<triton::usize, SolverModel>> allModels;
+
+        allModels = this->getModels(node, 1, status, timeout);
+        if (allModels.size() > 0)
+          ret = allModels.front();
```
SweetVishnya, line 140: return

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+        return ret;
```
SweetVishnya, line 142: return {}

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,182 @@
...
+          switch (res) {
+            case BITWUZLA_SAT:
+              *status = triton::engines::solver::SAT;
+              break;
+            case BITWUZLA_UNSAT:
+              *status = triton::engines::solver::UNSAT;
+              break;
+            case BITWUZLA_UNKNOWN:
+              if (code == 1)
+                *status = triton::engines::solver::TIMEOUT;
```
SweetVishnya, line 157: TODO: OUTOFMEM

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,182 @@
...
+        this->memoryLimit = 0;
+      }
+
+      int32_t term_fun(void* state) {
+        auto p = reinterpret_cast<SolverParams*>(state);
+        if (!p->timeout || ++p->counter < 1000) {
+          return 0;
+        }
+        p->counter = 0;
+        std::chrono::duration<double> delta = std::chrono::system_clock::now() - p->start;
```
SweetVishnya, line 35: std::chrono::duration_cast<std::chrono::milliseconds>

SweetVishnya (changes requested):

cnheitman:
> I am not able to build Triton with only -DBITWUZLA_INTERFACE=On: it requires Bitwuzla’ include directory to compile. So I think it can be organized the same way as z3 - both via find_package (fixed up to Bitwuzla) and INCLUDE_DIRS/INCLUDE_LIBRARIES variables.

Ok, let's maintain the `INCLUDE_DIRS` and `INCLUDE_LIBRARIES` variables then. 

BTW, thanks @apach301 for checking and clarifying the issue with the `forall` node.

> @cnheitman maybe we can rename evaluateAstViaZ3 by evaluateAstViaSolver if Bitwuzla allow us to evaluate an AST to a concrete value if there is no sym var? (specific z3 tests is also ok for me, but may be sexier with evaluateAstViaSolver)

@JonathanSalwan I like your proposition to rename `evaluateAstViaZ3` to `evaluateAstViaSolver`. Currently, there is no `evaluate` method in `BitwuzlaSolver`, perhaps we can add it before merging it. @apach301 Do you think it would easy to implement it? 

> @apach301, I think if the support of quantifier is not reliable yet on Bitwuzla, we can simply skip it and raise an exception like below:

Yes, I agree, we can skip it and raise an exception. We should skip the tests that use it as well (there only one right now), I'll find a way to do it.

JonathanSalwan:
With the commit 610e41cece2a8ffb6b09dbf33f706ea62b986ea3, i prepared the Python bindings part. I've:

* Renamed the SOLVER namespace to SOLVER_STATE (OUTOFMEM, SAT, etc.)
* Added the SOLVER.Z3 enum
* Exported the `TritonContext::setSolver` method
* Add an unittest

Now we just have to export `SOLVER.BITWUZLA` into the namespace when it's ready and update the unittest to test the solver switching.

cnheitman:
Nice @JonathanSalwan :) I was working on the same thing yesterday (bindings for `setSolver` and `getSolver`), but I was waiting until we merge to commit (we overlapped a bit there :)). 

In my case I added the following code to `synthetizing_obfuscated_expressions.py` as well, this forces to use Z3 (and avoids tests errors in case of using just Bitwuzla):

```diff
diff --git a/src/examples/python/synthetizing_obfuscated_expressions.py b/src/examples/python/synthetizing_obfuscated_expressions.py
index 4c79b181..9d2bc22c 100755
--- a/src/examples/python/synthetizing_obfuscated_expressions.py
+++ b/src/examples/python/synthetizing_obfuscated_expressions.py
@@ -157,6 +157,13 @@ def x_xor_92_obfuscated(x):
 
 def main():
     ctx = TritonContext(ARCH.X86_64)
+
+    try:
+        ctx.setSolver(SOLVER.ENGINE.SOLVER_Z3)
+    except:
+        print("This script requires Z3 as the solver engine. Compile Triton with Z3 support and re-run.")
+        sys.exit(0)
+
     ast = ctx.getAstContext()
 
     ctx.setAstRepresentationMode(AST_REPRESENTATION.PYTHON)
```

JonathanSalwan:
Hey @cnheitman, yeah I think I will stop commiting to dev-v0.9 to avoid rebase as much as possible for @apach301 :).

src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
```diff
@@ -139,14 +139,19 @@ namespace triton {
...
-          auto sort = bv_sorts.find(size);
+          auto bv_size = reinterpret_cast<IntegerNode*>(childNodes[1].get())->getInteger().convert_to<size_t>();
+          auto sort = bv_sorts.find(bv_size);
           if (sort == bv_sorts.end()) {
-            sort = bv_sorts.insert(std::make_pair(size, bitwuzla_mk_bv_sort(bzla, size))).first;
+            sort = bv_sorts.insert({bv_size, bitwuzla_mk_bv_sort(bzla, bv_size)}).first;
+          }
+
+          // Handle bitvector value as integer if it small enough.
+          if (bv_size <= sizeof(uint64_t) * 8) {
```
JonathanSalwan, line 149: I would have said `sizeof(void*)`? (what if the host is 32bit?)

SweetVishnya: Bitwuzla [takes](https://github.com/bitwuzla/bitwuzla/blob/76865c90622e29e7ff94bb8725a0ed8f4e35332e/src/api/c/bitwuzla.h#L2615) uint64_t on both architectures.

src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
```diff
@@ -204,10 +208,11 @@ namespace triton {
         case STRING_NODE: {
           std::string value = reinterpret_cast<triton::ast::StringNode*>(node.get())->getString();
 
-          if (symbols.find(value) == symbols.end())
+          const auto& it = symbols.find(value);
```
SweetVishnya, line 211: auto it =

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -110,36 +117,24 @@ namespace triton {
...
-        allModels = this->getModels(node, 1, status, timeout);
-        if (allModels.size() > 0)
-          ret = allModels.front();
+        auto allModels = this->getModels(node, 1, status, timeout);
+        if (!allModels.empty()) {
+          return allModels.front();
+        }
 
-        return ret;
+        return {};
```
SweetVishnya, line 137: return ternary operator

SweetVishnya (changes requested):

SweetVishnya:
One more TODO:

We should rebase and support solving time measurement as in #1064

SweetVishnya:
Useful Bitwuzla setup advises can be found in this thread https://github.com/bitwuzla/bitwuzla/issues/11

src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
apach301: @JonathanSalwan I think it should work on 32bit host too because it have the same data type.

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,243 @@
...
+          return 1;
+        }
+
+        // Check memory limit.
+        delta /= 1000;
+        if (p->memory_limit && delta > p->last_mem_check && delta % p->delay == 0) {
+          p->last_mem_check = delta;
+
+          size_t memory_usage = 0;
+#if defined(__unix__)
```
SweetVishnya, line 50: all memory_limit code can be under define (not implemented for Windows)

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+        // Check timeout expired.
+        if (p->timeout && delta > p->timeout) {
+          p->code = 1;
+          return 1;
+        }
+
+        // Check memory limit.
+        delta /= 1000;
+        if (p->memory_limit && delta > p->last_mem_check && delta % p->delay == 0) {
```
SweetVishnya, line 46: delta > p->last_mem_check is signed/unsigned comparison? maybe better check delta != last_mem_check

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,243 @@
...
+          }
+#endif
+          if (!memory_usage) {
+            return 0;
+          }
+          if (memory_usage > p->memory_limit) {
+            p->code = 2;
+            return 2;
+          }
+          if (memory_usage < 0.25 * p->memory_limit) {
```
SweetVishnya, line 71: p->memory_limit / 4

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,243 @@
...
+            p->delay = 5;
+          }
+          else if (memory_usage < 0.75 * p->memory_limit) {
```
SweetVishnya, line 74: 3 * p->memory_limit / 4

SweetVishnya (changes requested):

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
apach301: it is signed comparison to make first check more convenient

SweetVishnya:
LGTM

SweetVishnya (approved):

SweetVishnya:
I suppose we can merge it now. The future roadmap is following:

1. @cnheitman, @JonathanSalwan, could you prepare CI build + tests for Bitwuzla on your side?
2. Python bindings should be working now, but they require some testing.
3. @apach301, will play with Bitwuzla configs, benchmark them, and prepare a separate PR.

src/libtriton/bindings/python/namespaces/initSolverNamespace.cpp
```diff
@@ -38,7 +38,11 @@ namespace triton {
       void initSolverNamespace(PyObject* solverDict) {
         PyDict_Clear(solverDict);
 
+        #if defined(TRITON_Z3_INTERFACE)
         xPyDict_SetItemString(solverDict, "Z3", PyLong_FromUint32(triton::engines::solver::SOLVER_Z3));
+        #elif defined(TRITON_BITWUZLA_INTERFACE)
+        xPyDict_SetItemString(solverDict, "BITWUZLA", PyLong_FromUint32(triton::engines::solver::SOLVER_BITWUZLA));
+        #endif
```
JonathanSalwan, line 45: I'm not sure it's an `elif` here. What if z3 and Bitwuzla are both compiled ? Should not be two `ifdef` ?

JonathanSalwan:
I will review the code in detail this week before merging.

JonathanSalwan:
I've forked the branch `dev-v0.9` to a new branch `dev-bitwuzla`, once the review will be done we will merge the MR to the  `dev-bitwuzla` branch. Thus we can work on `CI-build` and unit tests on our side and close this MR.

cnheitman:
I will also take a look at this as soon as I can. I think it's a good idea to work on the `dev-bitwuzla` branch until we get the CI and tests working properly.

SweetVishnya:
Playing with Bitwuzla configs didn't help with performance. [These](https://github.com/bitwuzla/bitwuzla/blob/main/src/parser/bzlasmt2.c#L471) options actually enable additional propagation solver call (aka Triton AST simplifications) before real solving. We have quite optimized ASTs already, so it just spends more time on preprocessing.

src/libtriton/bindings/python/namespaces/initSolverNamespace.cpp
apach301: Fixed that bug

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+        // Check timeout expired.
+        if (p->timeout && delta > p->timeout) {
+          p->code = 1;
+          return 1;
+        }
+
+#if defined(__unix__)
+        // Check memory limit.
+        delta /= 1000;
```
JonathanSalwan, line 46: Can you add a comment about the constant `1000`. Where it comes from?

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,245 @@
...
+          if (!memory_usage) {
+            return 0;
+          }
+          if (memory_usage > p->memory_limit) {
+            p->code = 2;
+            return 2;
+          }
+          // The more memory is available the rarely we check current
+          // memory consumption in order to save resources.
+          if (memory_usage < p->memory_limit / 4) {
```
JonathanSalwan, line 72: Can you add a comment about the constant 4? Where it comes from?

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,245 @@
...
+            p->delay = 5;
+          }
+          else if (memory_usage < 3 * p->memory_limit / 4) {
```
JonathanSalwan, line 75: Can you add a comment about the constant 3 and 4? Where they come from?

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,245 @@
...
+            p->delay = 2;
+          }
+          else {
+            p->delay = 1;
+          }
+        }
+#endif
+
+        return 0;
+      }
```
JonathanSalwan, line 85: Regarding this whole function. You return either 0, 1 or 2. It would not be better to use enums?

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+        std::vector<std::unordered_map<triton::usize, SolverModel>> ret;
+        while(res == BITWUZLA_SAT && limit >= 1) {
+          std::vector<const BitwuzlaTerm*> solution;
+          solution.reserve(bzlaAst.getVariables().size());
+
+          // Parse model.
+          std::unordered_map<triton::usize, SolverModel> model;
+          for (const auto& it : bzlaAst.getVariables()) {
+            const char *svalue = bitwuzla_get_bv_value(bzla, it.first);
```
JonathanSalwan, line 140: `const char* svalue`

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,245 @@
...
+            triton::uint512 value = strtoull(svalue, 0, 2L);
+            auto m = SolverModel(it.second, value);
+            model[m.getId()] = m;
+
+            const auto& symvar_sort = bzlaAst.getBitvectorSorts().at(it.second->getSize());
+            auto cur_val = bitwuzla_mk_bv_value(bzla, symvar_sort, svalue, BITWUZLA_BV_BASE_BIN);
+            auto n = bitwuzla_mk_term2(bzla, BITWUZLA_KIND_EQUAL, it.first, cur_val);
+            solution.push_back(bitwuzla_mk_term1(bzla, BITWUZLA_KIND_NOT, n));
```
JonathanSalwan, line 148: Can you add a comment about what is does with the 4 above lines?

src/libtriton/includes/triton/bitwuzlaSolver.hpp
```diff
@@ -0,0 +1,125 @@
...
+            int64_t timeout;             /*!< Timeout (ms) for solver instance running. */
+            size_t  memory_limit;        /*!< Memory limit for the whole symbolic process. */
+            size_t  call_cnt = 0;        /*!< Counter for the number of termination callback calls. */
+            int64_t last_mem_check = -1; /*!< Time when the last memory usage check was performed. */
+            size_t  delay = 1;           /*!< Check memory limit every delay seconds. */
+            size_t  code = 0;            /*!< Reason of solving termination (1 - timeout, 2 - out_of_mem). */
+          };
+
+        //! Callback function that implements termination of Bitwuzla solver on timeout and memory limit.
+        static int32_t terminate_fun(void* state);
```
JonathanSalwan, line 68: Is static needed (i think yes but just to be sure)? Maybe also renamed it to `terminate` or `terminateCallback` (to keep the correct coding style)?

src/libtriton/includes/triton/bitwuzlaSolver.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+        //! Callback function that implements aborting of Bitwuzla solver with throwing exception.
+        static void abort_fun(const char* msg);
```
JonathanSalwan, line 71: Is static needed (i think yes but just to be sure)? Maybe also renamed it to `abort` or `abortCallback` (to keep the correct coding style)?

src/libtriton/includes/triton/tritonToBitwuzlaAst.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+        //! Destructor.
+        TRITON_EXPORT ~TritonToBitwuzlaAst();
+
+        //! Converts to Bitwuzla's AST
+        TRITON_EXPORT const BitwuzlaTerm* convert(const SharedAbstractNode& node, Bitwuzla* bzla);
+
+        //! Returns symbolic variables and its assosiated Bitwuzla terms to process the solver model.
+        TRITON_EXPORT const std::unordered_map<const BitwuzlaTerm*, triton::engines::symbolic::SharedSymbolicVariable>& getVariables() const {
+          return variables;
```
JonathanSalwan, line 52: Can you avoid as much as possible to write code in header?

src/libtriton/includes/triton/tritonToBitwuzlaAst.hpp
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        //! Returns bitvector sorts.
+        TRITON_EXPORT const std::map<size_t, const BitwuzlaSort*>& getBitvectorSorts() const {
+          return bv_sorts;
```
JonathanSalwan, line 57: Can you avoid as much as possible to write code in header?

src/libtriton/includes/triton/tritonToBitwuzlaAst.hpp
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+      private:
+        //! The convert internal process.
+        const BitwuzlaTerm* translate(const SharedAbstractNode& node, Bitwuzla* bzla);
+
+        //! The map of Triton's AST nodes translated to the Bitwuzla terms.
+        std::unordered_map<SharedAbstractNode, const BitwuzlaTerm*> translated_nodes;
```
JonathanSalwan, line 65: `translatedNodes`

src/libtriton/includes/triton/tritonToBitwuzlaAst.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+        //! The set of symbolic variables contained in the expression.
+        std::unordered_map<const BitwuzlaTerm*, triton::engines::symbolic::SharedSymbolicVariable> variables;
+
+        //! The map of symbols. E.g: (let (symbols expr1) expr2)
+        std::unordered_map<std::string, triton::ast::SharedAbstractNode> symbols;
+
+        //! All bitvector sorts that used in the expression.
+        std::map<size_t, const BitwuzlaSort*> bv_sorts;
```
JonathanSalwan, line 74: `bvSorts`

JonathanSalwan:
Looks good to me. Just minor renaming or adding comment here and there.

cnheitman (approved):

cnheitman:
It looks good to me too! I agree on the renaming and commenting that @JonathanSalwan mentioned.

Really nice PR @apach301 ! :rocket:  

Hey @JonathanSalwan , regarding the items that are pending, how do you want to split them? You handle the CI and
I add some more tests?

For the CI, should we test separate builds for Z3 and Bitwusla? (I think that would be the best approach.)

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
SweetVishnya: We may use the already existing enum for solver status.

JonathanSalwan:
> Hey @JonathanSalwan , regarding the items that are pending, how do you want to split them? You handle the CI and
I add some more tests?

Yep, we can do that :)

> For the CI, should we test separate builds for Z3 and Bitwusla? (I think that would be the best approach.)

I think it's the best approach too. We can add two more github actions with the following build confs:

* Z3=on and Bitwusla=on (so that we can switch from a solver to another during tests)
* Z3=off and Bitwusla=on (just to see if only with Bitwusla all unit tests are passing)

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
apach301: Terminate function should return non-zero to terminate solving, zero otherwise. Refactored callback to return 0/1 and to use status_e for the reason of termination (`SolverParams.code`)

JonathanSalwan:
Thx @apach301 for the last commit. I will merge everything into the `dev-bitwuzla` branch.

Congrats **everyone** for this MR! :rocket:

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+          size_t memory_usage = 0;
+          std::ifstream str("/proc/self/status");
+          if (str.is_open()) {
+            std::string line;
+            std::regex vmsize("VmRSS:\\s*([0-9]+)\\s*kB");
+            while (std::getline(str, line)) {
+              std::smatch match;
+              if (std::regex_match(line, match, vmsize)) {
+                memory_usage = strtoul(match.str(1).c_str(), NULL, 10) / 1024;
```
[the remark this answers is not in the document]
apach301: In this case `strtoul` shouldn't fail due to regex check, which ensures that line have at least 1 digit.

apach301:
About implementing `evaluateAstViaSolver` for Bitwuzla. It have [API](https://bitwuzla.github.io/docs/c/interface.html#c.bitwuzla_simplify) for simplifying the formula, but it returns only result (SAT/UNSAT/UNKNOWN). As much as I understand, Bitwuzla can simplify formula only to true/false. Maybe there is some way to get evaluated number, but I can't find it quickly.

apach301:
When quantifiers will be supported in Bitwuzla it will be possible to implement `forall` node in tritonToBitwuzlaAst.cpp:
```
case FORALL_NODE: {
  // throw triton::exceptions::SolverEngine("forall not supported.");
  std::vector<const BitwuzlaTerm*> args;
  std::vector<const BitwuzlaTerm*> keys;
  std::vector<const BitwuzlaTerm*> values;

  for (const auto& child : node->getChildren()) {
    if (child->getType() == VARIABLE_NODE) {
      auto const_node = translated_nodes.at(child);
      auto var_node = variableNodes.at(const_node);
      keys.push_back(const_node);
      values.push_back(var_node);
      args.push_back(var_node);
      continue;
    }
    auto n = translated_nodes.at(child);
    args.push_back(bitwuzla_substitute_term(bzla, n, keys.size(), keys.data(), values.data()));
  }
  return bitwuzla_mk_term(bzla, BITWUZLA_KIND_FORALL, args.size(), args.data());
}
```
This solution also requires creating both `bitwuzla_mk_const` and `bitwuzla_mk_var` for every `VARIABLE_NODE`.

JonathanSalwan:
@cnheitman, should be ok with the CI of Z3 and Bitwuzla.

I've pushed a [simple test](https://github.com/JonathanSalwan/Triton/blob/c30a3c358fe0159708cfde68689b19be6b308f65/src/testers/unittests/test_solver.py) to verify if we can [switch](https://github.com/JonathanSalwan/Triton/blob/c30a3c358fe0159708cfde68689b19be6b308f65/src/testers/unittests/test_solver.py#L33-L38) to a solver from another and get the [same result](https://github.com/JonathanSalwan/Triton/blob/c30a3c358fe0159708cfde68689b19be6b308f65/src/testers/unittests/test_solver.py#L26-L28). We can also see that some parts of the Bitwuzla's code has been [covered](https://app.codecov.io/gh/JonathanSalwan/Triton/blob/c30a3c358fe0159708cfde68689b19be6b308f65/src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp) and what else need to be covered for our future unit tests.

PR #1066, opened 2021-12-29
Fix double z3 solver check call

SweetVishnya:
We found out that z3 check was called twice. Seems like an easy speedup :) Thanks to @apach301

### DIFF
--- a/src/libtriton/engines/solver/z3/z3Solver.cpp
+++ b/src/libtriton/engines/solver/z3/z3Solver.cpp
@@ -133,7 +133,9 @@ namespace triton {
 
             if (--limit) {
               /* Escape last models */
-              solver.add(triton::engines::solver::mk_or(args));
+              if (!args.empty()) {
+                solver.add(triton::engines::solver::mk_or(args));
+              }
 
               /* Get next model */
               res = solver.check();
--- a/src/libtriton/engines/solver/z3/z3Solver.cpp
+++ b/src/libtriton/engines/solver/z3/z3Solver.cpp
@@ -22,8 +22,7 @@ namespace triton {
   namespace engines {
     namespace solver {
 
-      //! Wrapper to handle variadict number of arguments or'd togethers
-      z3::expr mk_or(z3::expr_vector args) {
+      z3::expr Z3Solver::mk_or(z3::expr_vector args) {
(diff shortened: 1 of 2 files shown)


src/libtriton/engines/solver/z3/z3Solver.cpp
```diff
@@ -123,18 +122,24 @@ namespace triton {
...
+            /* Push model */
+            ret.push_back(smodel);
 
-            /* If there is model available */
-            if (smodel.size() > 0)
-              ret.push_back(smodel);
+            if (--limit) {
+              /* Escape last models */
+              if (!args.empty()) {
+                solver.add(triton::engines::solver::mk_or(args));
```
JonathanSalwan, line 137: As we will probably have Bitwuzla into Triton soon, can you move `::solver::mk_or` into a new `z3` namespace like `::solver::z3::mk_or` ?

JonathanSalwan (changes requested):

JonathanSalwan:
Nice catch guys!

src/libtriton/engines/solver/z3/z3Solver.cpp
SweetVishnya: z3 already has its one `z3` namespace. I suppose this can mess up the namespaces:
```
/home/vishnya/fwork/sydr/triton/src/libtriton/engines/solver/z3/z3Solver.cpp:272:62: error: ‘check_result’ is not a member of ‘triton::engines::solver::z3’
       void Z3Solver::writeBackStatus(z3::solver& solver, z3::check_result res, triton::engines::solver::status_e* status) const {

```

SweetVishnya: I can make it a member function of Z3Solver

JonathanSalwan: Yes, a static method then?

JonathanSalwan: Well, indeed. If it's only used in `Z3Solver` we can add it as private method.

SweetVishnya: I pushed a fix.

PR #1068, opened 2022-01-14
Bitwuzla evaluate

apach301:
Implemented `evaluate()` method for Bitwuzla solver. Also renamed `evaluateAstViaZ3()` -> `evaluateAstViaSolver()`. Unittests from `test_ast_eval.py` now pass for both solvers.

### DIFF
--- a/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
+++ b/src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
@@ -18,7 +18,8 @@
 namespace triton {
   namespace ast {
 
-    TritonToBitwuzlaAst::TritonToBitwuzlaAst() {
+    TritonToBitwuzlaAst::TritonToBitwuzlaAst(bool eval)
+      : isEval(eval) {
     }
 
 
@@ -242,6 +243,16 @@ namespace triton {
           if (sort == bvSorts.end()) {
             sort = bvSorts.insert({size, bitwuzla_mk_bv_sort(bzla, size)}).first;
           }
+
+          // If the conversion is used to evaluate a node, we concretize symbolic variables.
+          if (this->isEval) {
+            triton::uint512 value = reinterpret_cast<triton::ast::VariableNode*>(node.get())->evaluate();
+            if (size <= sizeof(uint64_t) * 8) {
+              return bitwuzla_mk_bv_value_uint64(bzla, sort->second, value.convert_to<uint64_t>());
+            }
+            return bitwuzla_mk_bv_value(bzla, sort->second, value.convert_to<std::string>().c_str(), BITWUZLA_BV_BASE_DEC);
+          }
+
           auto n = bitwuzla_mk_const(bzla, sort->second, symVar->getName().c_str());
           variables[n] = symVar;
           return n;
--- a/src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
+++ b/src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
@@ -163,7 +163,7 @@ namespace triton {
           std::unordered_map<triton::usize, SolverModel> model;
           for (const auto& it : bzlaAst.getVariables()) {
             const char* svalue = bitwuzla_get_bv_value(bzla, it.first);
-            triton::uint512 value = strtoull(svalue, 0, 2L);
+            auto value = this->fromBvalueToUint512(svalue);
             auto m = SolverModel(it.second, value);
             model[m.getId()] = m;
 
@@ -242,27 +242,7 @@ namespace triton {
         // Evaluate concrete AST in solver.
         auto bzlaAst = triton::ast::TritonToBitwuzlaAst(true);
         auto bv_value = bitwuzla_get_bv_value(bzla, bitwuzla_get_value(bzla, bzlaAst.convert(node, bzla)));
-
-        // Convert short bitvector string directly.
-        triton::uint512 res;
-        auto len = strlen(bv_value);
-        if (len <= 64) {
-          res = strtoull(bv_value, 0, 2L);
-          bitwuzla_delete(bzla);
-          return res;
-        }
-
-        // Convert long bitvector string by 64-bit chunks.
-        uint pos = 0;
-        while (pos < len) {
-          auto sublen = std::min(len - pos, 64UL);
-          char substr[sublen + 1];
-          memcpy(substr, bv_value + pos, sublen);
-          substr[sublen] = '\0';
-          triton::uint512 value = strtoull(substr, 0, 2L);
-          res = (res << sublen) + value;
-          pos += sublen;
-        }
+        auto res = this->fromBvalueToUint512(bv_value);
 
         bitwuzla_delete(bzla);
         return res;
@@ -281,6 +261,28 @@ namespace triton {
 
       void BitwuzlaSolver::setMemoryLimit(triton::uint32 limit) {
         this->memoryLimit = limit;
+      }
+
+
+      triton::uint512 BitwuzlaSolver::fromBvalueToUint512(const char* value) const {
+        // Convert short bitvector string directly.
+        auto len = strlen(value);
+        if (len <= 64) {
+          return strtoull(value, 0, 2L);
+        }
+
+        // Convert long bitvector string by 64-bit chunks.
+        uint pos = 0;
+        triton::uint512 res;
+        while (pos < len) {
+          auto sublen = std::min(len - pos, 64UL);
+          char substr[sublen + 1];
+          memcpy(substr, value + pos, sublen);
+          substr[sublen] = '\0';
+          res = (res << sublen) + strtoull(substr, 0, 2L);
+          pos += sublen;
+        }
+        return res;
       }
 
     };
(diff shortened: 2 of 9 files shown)


SweetVishnya:
Are these evaluation tests running in CI?

JonathanSalwan:
Cool! Btw i think you forgot some debug info into `BitwuzlaSolver::terminateCallback`

SweetVishnya:
> Cool! Btw i think you forgot some debug info into `BitwuzlaSolver::terminateCallback`

Yeah, we should leave only `static` for faster regexp. Btw, we have some issues with proper timeout/oom handling. The callback may not be called for 20 seconds. @apach301 is going to figure it out in further PRs.

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -226,6 +226,49 @@ namespace triton {
...
+          res = strtoull(bv_value, 0, 2L);
+          bitwuzla_delete(bzla);
+          return res;
+        }
+
+        // Convert long bitvector string by 64-bit chunks.
+        uint pos = 0;
+        while (pos < len - 1) {
+          auto sublen = std::min(len - pos, 64UL);
+          char substr[sublen];
```
SweetVishnya, line 259: sublen + 1 (stack bof)

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -226,6 +226,49 @@ namespace triton {
...
+          memcpy(&substr, bv_value + pos, sublen);
```
SweetVishnya, line 260: memcpy(substr, bv_value + pos, sublen)

SweetVishnya (changes requested):

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -226,6 +226,49 @@ namespace triton {
...
+        auto len = strlen(bv_value);
+        if (len <= 64) {
+          res = strtoull(bv_value, 0, 2L);
+          bitwuzla_delete(bzla);
+          return res;
+        }
+
+        // Convert long bitvector string by 64-bit chunks.
+        uint pos = 0;
+        while (pos < len - 1) {
```
SweetVishnya, line 257: pos < len

SweetVishnya: corner case len == 65

src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
```diff
@@ -242,6 +243,14 @@ namespace triton {
           if (sort == bvSorts.end()) {
             sort = bvSorts.insert({size, bitwuzla_mk_bv_sort(bzla, size)}).first;
           }
+
+          // If the conversion is used to evaluate a node, we concretize symbolic variables.
+          if (this->isEval) {
+            triton::uint512 value = reinterpret_cast<triton::ast::VariableNode*>(node.get())->evaluate();
+            std::string bv_value(value.convert_to<std::string>());
+            return bitwuzla_mk_bv_value(bzla, sort->second, bv_value.c_str(), BITWUZLA_BV_BASE_DEC);
```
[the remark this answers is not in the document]
apach301: We concretize symbolic variable according to Triton AST, so parsing it value from Bitwuzla' model is pointless.

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -36,7 +36,7 @@ namespace triton {
         auto p = reinterpret_cast<SolverParams*>(state);
 
         // Execute this callback only once in every 1000 calls.
-        if (++p->call_cnt < 1000) {
+        if (++p->call_cnt < 100 && p->last_mem_check >= 0) {
```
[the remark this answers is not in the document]
apach301: Roll back this changes

src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
```diff
@@ -242,6 +243,14 @@ namespace triton {
           if (sort == bvSorts.end()) {
             sort = bvSorts.insert({size, bitwuzla_mk_bv_sort(bzla, size)}).first;
           }
+
+          // If the conversion is used to evaluate a node, we concretize symbolic variables.
+          if (this->isEval) {
+            triton::uint512 value = reinterpret_cast<triton::ast::VariableNode*>(node.get())->evaluate();
+            std::string bv_value(value.convert_to<std::string>());
```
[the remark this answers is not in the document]
JonathanSalwan: Does it really improve perf? because it's also nice to have short line of code for readability. Code is like art, must be beautiful to look :)

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
JonathanSalwan: `sublen - 1`?

src/libtriton/ast/bitwuzla/tritonToBitwuzlaAst.cpp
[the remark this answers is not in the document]
apach301: First, we skip conversion from integer to string when obtaining value from Triton node. Second, on Bitwuzla size we skip string validation (it have multiple checks for valid characters and width) + it converts it back from string to gmp::multiprecision integer. When we use `bitwuzla_mk_bv_value_uint64` it only converts from one multiprecision integer type to another.

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -226,6 +226,49 @@ namespace triton {
...
+        uint pos = 0;
+        while (pos < len) {
+          auto sublen = std::min(len - pos, 64UL);
+          char substr[sublen + 1];
+          memcpy(substr, bv_value + pos, sublen);
+          substr[sublen] = '\0';
+          triton::uint512 value = strtoull(substr, 0, 2L);
+          res = (res << sublen) + value;
+          pos += sublen;
+        }
```
JonathanSalwan, line 265: Can you extract those both parts of code into an outside private function called `fromBvalueToUint512()`? So that at the end we have:

```cpp
        auto bzlaAst = triton::ast::TritonToBitwuzlaAst(true);
        auto bv_value = bitwuzla_get_bv_value(bzla, bitwuzla_get_value(bzla, bzlaAst.convert(node, bzla)));
        triton::uint512 res = this->fromBvalueToUint512(bv_value);
```

JonathanSalwan:
Btw, when reading this patch I found [something](https://github.com/JonathanSalwan/Triton/blob/dev-bitwuzla/src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp#L166). Here you consider than symbolic variables will not be greater than 64-bits long?

apach301:
> Btw, when reading this patch I found something. Here you consider than symbolic variables will not be greater than 64-bits long?
Yeah, that's a bug) I'll change it to `fromBvalueToUint512()`. Also one more unittest on long symbolic variables needed.

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
apach301: Done

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
```diff
@@ -226,6 +226,29 @@ namespace triton {
...
+        if (node == nullptr) {
+          throw triton::exceptions::AstTranslations("BitwuzlaSolver::evaluate(): node cannot be null.");
+        }
+
+        // Perform check-sat on empty solver to enable model generation.
+        auto bzla = bitwuzla_new();
+        bitwuzla_set_option(bzla, BITWUZLA_OPT_PRODUCE_MODELS, 1);
+        if (bitwuzla_check_sat(bzla) != BITWUZLA_SAT) {
+          bitwuzla_delete(bzla);
+          return 0;
```
SweetVishnya, line 239: Is it a possible control flow or should we throw an exception here?

SweetVishnya (changes requested):

src/libtriton/engines/solver/bitwuzla/bitwuzlaSolver.cpp
JonathanSalwan: MMmmh that's a good question because for me it's cannot be unsat as there is no symbolic variable? There is no equation neither? For example we could be able to evaluate `2 + 3 * 4`, so there is no SAT reasoning here.

apach301: In this case we call `bitwuzla_check_sat()` on empty solver, so its always should be SAT. The concrete AST is translated and sent to solver later. Querying check-sat is necessary, because without it solver won't allow getting values from it.

JonathanSalwan: Can you add what you wrote into a comment to keep this understanding when reading code?

PR #1071, opened 2022-01-20
Bitwuzla termination callback

apach301:
- Throw exception when on Bitwuzla failure on `evaluate()`
- Termination callback refactoring: remove `call_cnt` (doesn't slow solving time notable), remove `delay`-logic (checking every second is enough), doesn't check oom in first 100ms of solving (no need to check memory on easy queries). Thus, termination callback become more concise and clear.


SweetVishnya (approved):
LGTM

PR #1088, opened 2022-02-16
[cmake] Fix pre-definintion and dynamic library installation, add options for examples and tests

JackBoosY:
1. `TRITON_BUILDING_DLL` is not exists, it should be `BUILDING_DLL`. See https://github.com/JonathanSalwan/Triton/blob/96104b1a860dc7ddb9ab123859b2fd668e388f72/src/libtriton/includes/triton/dllexport.hpp#L12-L17
2. `BUILDING_DLL` should only be used in build process, so it must be `PRIVATE`, not `PUBLIC`.
3. The dynamic library is not installed.
4. Add options for feature `examples` and `tests`.

Related: https://github.com/microsoft/vcpkg/pull/23111

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,8 @@ option(GCOV                 "Enable code coverage"                     OFF)
 option(LLVM_INTERFACE       "Use LLVM for lifting"                     OFF)
 option(MSVC_STATIC          "Use statically-linked runtime library"    OFF)
 option(Z3_INTERFACE         "Use Z3 as SMT solver"                     ON)
+option(BUILD_EXAMPLES       "Build the examples"                       ON)
+option(ENABLE_TEST          "Build test program"                       ON)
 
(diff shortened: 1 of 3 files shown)


JackBoosY:
Wow, so quickly.

JonathanSalwan:
Btw @JackBoosY, regarding https://github.com/microsoft/vcpkg/pull/23111#issuecomment-1041270332 what fix are you talking about?

JackBoosY:
> Btw @JackBoosY, regarding [microsoft/vcpkg#23111 (comment)](https://github.com/microsoft/vcpkg/pull/23111#issuecomment-1041270332) what fix are you talking about?

Ah that's only fits for vcpkg, we don't need to fix that here.

JackBoosY:
And there is another build issue: https://github.com/JonathanSalwan/Triton/issues/1089

PR #1094, opened 2022-02-17
Dockerfile for Triton with LLVM, Z3, and bitwuzla

packmad:
Triton fully-equipped: LLVM for lifting; z3 or bitwuzla as SMT solver

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,30 +1,39 @@
 FROM ubuntu:20.04
-LABEL maintainer="David Manouchehri"
+ARG DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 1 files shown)


packmad:
@JonathanSalwan sorry for the double PR, anyway, this Dockerfile has everything :)

JonathanSalwan:
Hey @packmad, we have an issue. With recent commit, we need to support cmake 3.20. Do you think it's possible to upgrade the ubuntu base image or at least install cmake 3.20 from source?

mrexodia:
You can do pip install cmake to install the latest

PR #1104, opened 2022-03-02
Use all cores to build in Windows

illera88:


src/libtriton/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Windows")
 endif()
 
 if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
+    add_definitions(/MP)
```
mrexodia, line 12: In general you shouldn't use commands to add compile options, especially because you don't know what compiler is used. `add_definitions` is a deprecated command used to add `#define` entries. If you want to compile with multiple cores you should pass the `/MP` flag to your CMake command line with `-DCMAKE_CXX_FLAGS=your custom flags`

mrexodia (changes requested):

src/libtriton/CMakeLists.txt
JonathanSalwan: Yeah i agree with what @mrexodia said.

PR #1108, opened 2022-03-14
Fix reference count of Py_True/Py_False

HyperSine:
In the end of `xPyDict_SetItemString`, it will call `Py_DECREF(val)`, which means this function treats `val` as strong reference. 

`Py_True` and `Py_False` are just constant pointers to `Py_Object`, so we must increase their reference count before passing them to `xPyDict_SetItemString`.


JonathanSalwan:
Nice catch. Thx!

PR #1116, opened 2022-04-06
Support Z3 package with imported targets

theblackunknown:
## Description

This PR improves the CMake setup for the `Z3` whether it found using the embedded `FindZ3.cmake` or pull through an installed CMake config [1].
Moreover, I have added version extraction from header to get a log during configure to know which Z3 version is being used.
I have also added a `libz3` names when searching for lib as it seems to be the default file name when installing it (cf. https://github.com/Z3Prover/z3/blob/08294d62e5cd7898d3876038264765bcc0c9d8e6/src/CMakeLists.txt#L118)

[1]  [Z3Prover/z3](https://github.com/Z3Prover/z3) is producing a CMake config described here (https://github.com/Z3Prover/z3/blob/0fa0feb97930a72b550bf0032e57a0ec8d3df63c/CMakeLists.txt#L544-L547)
This is being used when we want to build and install Triton through a package manager tool like [microsoft/vcpkg](https://github.com/microsoft/vcpkg)


CMakeModules/FindZ3.cmake
```diff
@@ -50,3 +51,20 @@ else()
     message(STATUS "Z3 includes directory defined: ${Z3_INCLUDE_DIRS}")
     message(STATUS "Z3 libraries defined: ${Z3_LIBRARIES}")
 endif()
+
+find_file(Z3_VERSION_HEADER
+  z3_version.h
+  PATHS ${Z3_PKGCONF_INCLUDE_DIRS}
```
SweetVishnya, line 57: This variable may not be present when building with z3 in custom directory that is not installed system-wide.

SweetVishnya: We may try searching in Z3_INCLUDE_DIRS if Z3_PKGCONF_INCLUDE_DIRS are not present.

mrexodia:
> > Just delete the FindZ3.cmake that ships with Triton and everything will work properly no?
> 
> We still want to support specifying custom z3 location via `-DZ3_INCLUDE_DIRS` and `-DZ3_LIBRARIES`.

Why would you want to support that? The user can specify their own FindZ3.cmake with whatever hacks they want, but Z3 ships with an official module…

JonathanSalwan:
> Just delete the FindZ3.cmake that ships with Triton and everything will work properly no?

This is working only if the user uses the cmake build of Z3. Their are two different ways to build z3 (`cmake` and `mk_make.py`). `mk_make.py` does not generate the `Z3Config.cmake` file.

CMakeModules/FindZ3.cmake
theblackunknown: I guess I have the same question as in #1117 : I am curious about the workflow you described.
So far I am aware of `find_package`:
- using the `MODULE` mode which uses `FindZ3.cmake`
- using the `CONFIG` mode which will uses a `Z3Config.cmake`, and then the way to define `Z3_VERSION` is document through the CMake Config mode mechanism.

theblackunknown:
My goal with this PR is to support both finding `Z3` whether we are in `MODULE` and `CONFIG` mode.
So I am willing to improve this PR to support anything not covered yet, feel free to submit a review so that I know which parts looks problematic like @SweetVishnya did

mrexodia:
> > > > Just delete the FindZ3.cmake that ships with Triton and everything will work properly no?
> > > 
> > > 
> > > We still want to support specifying custom z3 location via `-DZ3_INCLUDE_DIRS` and `-DZ3_LIBRARIES`.
> > 
> > 
> > Why would you want to support that? The user can specify their own FindZ3.cmake with whatever hacks they want, but Z3 ships with an official module…
> 
> I just want to git clone z3 somewhere, then build it in some directory and link Triton with it.

I think this is a misunderstanding about how CMake works. You can do exactly what you want with `find_package`:

```sh
# Clone a specific z3 tag
git clone https://github.com/Z3Prover/z3
cd z3
git checkout z3-4.8.15

# Build z3
cmake -G Ninja -B build -DCMAKE_BUILD_TYPE=Release
cmake --build build

# Install z3 locally
cmake --install build --prefix install

# Now from the Triton folder
cd ..\Triton
cmake -B build -DCMAKE_PREFIX_PATH=$(pwd)/../z3/install
```

This way `find_package(Z3 REQUIRED)` will find your locally cloned Z3. This is the intended way for CMake to consume external packages. The created `install` folder is a portable installation and you can ZIP it, cache it on CI etc and it will work.

theblackunknown:
@SweetVishnya my latest pushed changes have solved the problem you raised back then.
Can you give it one more try and let me know if this PR is in a good shape ?

CMakeLists.txt
```diff
@@ -104,7 +104,14 @@ list(APPEND CMAKE_MODULE_PATH "${TRITON_ROOT}/CMakeModules/")
 if(Z3_INTERFACE)
     message(STATUS "Compiling with Z3 SMT solver")
     find_package(Z3 REQUIRED)
-    include_directories(${Z3_INCLUDE_DIRS})
+    message(STATUS "Z3 version: ${Z3_VERSION}")
+    if(TARGET z3::libz3)
+        link_libraries(z3::libz3)
+    elseif(DEFINED Z3_INCLUDE_DIRS)
+        include_directories(${Z3_INCLUDE_DIRS})
```
JonathanSalwan, line 111: I do not understand why it's an `elseif` there? `link_libraries` and `include_directories` have not the same behavior?

SweetVishnya:
> @SweetVishnya my latest pushed changes have solved the problem you raised back then. Can you give it one more try and let me know if this PR is in a good shape ?

Both #1116 and #1117 work for me. However, I am not able to verify the find package way.

CMakeLists.txt
theblackunknown: Not in this case `z3::libz3` is a CMake target describing _how_ to link to it, meaning it describes what are the include directories to include and which libraries to link to.
The way to leverages those description from a CMake target is by using `(target_)link_libraries` in CMake.

theblackunknown: `include_directories` CMake command expected path which will simply be added to the compiler command line accordingly, there is no "smart resolution" like when using `(target_)link_librarie` with CMake targets

theblackunknown:
> > @SweetVishnya my latest pushed changes have solved the problem you raised back then. Can you give it one more try and let me know if this PR is in a good shape ?
> 
> Both #1116 and #1117 work for me. However, I am not able to verify the find package way.

Thanks for testing !

I have tested those on my end using `vcpkg`, I can give direction on how to test those locally with it.
Alternatively, I have also designed a custom Github Action workflow which illustrates, are you interested that I also contributes it to the repo ?

JonathanSalwan:
I'm merging this one and the one for capstone. @theblackunknown can you open a new MR for the github action and vcpkg?

theblackunknown:
> I'm merging this one and the one for capstone. @theblackunknown can you open a new MR for the github action and vcpkg?

Sure ! Will do tomorrow 👍

Issue #972, opened 2020-11-26
capstone-next: x86Specifications.cpp:156:43: error: 'X86_INS_FADDP' is not a member of 'triton::extlibs::capstone'; did you mean 'X86_INS_FADD'?   156 |           case triton::extlibs::capstone::X86_INS_FADDP:


XVilka:
Trying to build latest Triton 0.9 branch (https://github.com/JonathanSalwan/Triton/commit/0e9fc8ea0a1c4d9d38ee09e80f926868ec96a34f) with the latest capstone 'next' (https://github.com/aquynh/capstone/commits/next) produces this error:
```
//Triton/src/Triton/src/libtriton/arch/x86/x86Specifications.cpp:156:43: error: 'X86_INS_FADDP' is not a member of 'triton::extlibs::capstone'; did you mean 'X86_INS_FADD'?
  156 |           case triton::extlibs::capstone::X86_INS_FADDP:
      |                                           ^~~~~~~~~~~~~
      |                                           X86_INS_FADD
/Triton/src/Triton/src/libtriton/arch/x86/x86Specifications.cpp:2880:43: error: 'X86_INS_UD2B' is not a member of 'triton::extlibs::capstone'; did you mean 'X86_INS_UD2'?
 2880 |           case triton::extlibs::capstone::X86_INS_UD2B:
      |                                           ^~~~~~~~~~~~
      |                                           X86_INS_UD2
//Triton/src/Triton/src/libtriton/arch/x86/x86Specifications.cpp:3040:43: error: 'X86_INS_VCVTPD2DQX' is not a member of 'triton::extlibs::capstone'; did you mean 'X86_INS_VCVTPD2DQ'?
 3040 |           case triton::extlibs::capstone::X86_INS_VCVTPD2DQX:
      |                                           ^~~~~~~~~~~~~~~~~~
      |                                           X86_INS_VCVTPD2DQ
/Triton/src/Triton/src/libtriton/arch/x86/x86Specifications.cpp:3048:43: error: 'X86_INS_VCVTPD2PSX' is not a member of 'triton::extlibs::capstone'; did you mean 'X86_INS_VCVTPD2PS'?
 3048 |           case triton::extlibs::capstone::X86_INS_VCVTPD2PSX:
      |                                           ^~~~~~~~~~~~~~~~~~
      |                                           X86_INS_VCVTPD2PS
[...]

PR #1118, opened 2022-04-06
Code changes to support using capstone next

theblackunknown:
## Description

This PR introduce code changes to allow using Triton with the current HEAD of [capstone-engine/capstone](https://github.com/capstone-engine/capstone) (as of writing).

I have seen the previous conversion in #972 but although capstone is not finalized yet I think there is some value to be able to try Triton with the current repository.
Especially as those code changes looks fairly backward compatible to my untrained eyes (disclaimer: I am not a heavy Triton user).

There is an interesting thread at capstone-engine/capstone#1456 which give many information about some of the capstone 5 changes, which I used as inspiration for this PR.

### DIFF
--- a/src/libtriton/arch/x86/x86Specifications.cpp
+++ b/src/libtriton/arch/x86/x86Specifications.cpp
@@ -155,9 +155,11 @@ namespace triton {
             tritonId = triton::arch::x86::ID_INS_FIADD;
             break;
 
+#if CS_API_MAJOR < 5
           case triton::extlibs::capstone::X86_INS_FADDP:
             tritonId = triton::arch::x86::ID_INS_FADDP;
             break;
+#endif
 
           case triton::extlibs::capstone::X86_INS_ADOX:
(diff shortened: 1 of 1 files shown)


JonathanSalwan:
Thanks for this PR. I still have to do more tests because last time I tried to use capstone v5, i got more issues than just enums. For example, some of instructions did not have the same information decoding from v4 and v5.

JonathanSalwan:
Thanks for this PR. I still have to do more tests because last time I tried to use capstone v5, i got more issues than just enums. For example, some of instructions did not have the same information decoding from v4 and v5.

theblackunknown:
> Thanks for this PR. I still have to do more tests because last time I tried to use capstone v5, i got more issues than just enums. For example, some of instructions did not have the same information decoding from v4 and v5.

Great ! I do not have a good knowledge of `capstone` but I am interested to hear more about those issues.
We should even consider reporting them as they have started to produce release candidates tags.

PR #1134, opened 2022-06-01
Support instruction assembly printing for symbolic expressions in lifting engine

### LINKED ISSUE #1133: Instruction printing in SMT dump is broken
Hi!

It seems that new smt lifting broke code from #982 that was printing assembly for SMT formulas.

SweetVishnya:
This PR brings back #982 for new lifting engine. One is able to use boolean argument to enable instruction assembly printing for symbolic expressions.

Closes #1133

### DIFF
--- a/src/libtriton/engines/lifters/liftingToPython.cpp
+++ b/src/libtriton/engines/lifters/liftingToPython.cpp
@@ -104,8 +104,10 @@ namespace triton {
           auto& e = ssa[id];
           stream << e->getFormattedExpression();
           if (icomment && !e->getDisassembly().empty()) {
-            if (!e->getComment().empty())
-            {
+            if (e->getComment().empty()) {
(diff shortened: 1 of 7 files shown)


src/libtriton/engines/lifters/liftingToPython.cpp
```diff
@@ -101,7 +101,16 @@ namespace triton {
 
         /* Print symbolic expressions */
         for (const auto& id : symExprs) {
-          stream << ssa[id]->getFormattedExpression() << std::endl;
+          auto& e = ssa[id];
+          stream << e->getFormattedExpression();
+          if (icomment && !e->getDisassembly().empty()) {
+            if (!e->getComment().empty())
```
SweetVishnya, line 107: Create comment if it is empty

PR #1137, opened 2022-06-11
more appveyor python versions?

brandonros:
not sure if we need to go excessive and add 3.8 + 3.9 + 3.10 for win32 too instead of just x64?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -16,6 +16,7 @@ environment:
   CMAKE_TEMPLATE: Visual Studio 16 2019
 
   matrix:
+    # 32-bit, python3.6, boost on
     - platform: Win32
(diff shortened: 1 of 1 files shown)


brandonros:
I'll just use Docker I guess, unless you think this would help other. Python3.6 + Python3.7 are pretty old from what I can tell?

JonathanSalwan:
> not sure if we need to go excessive and add 3.8 + 3.9 + 3.10 for win32 too instead of just x64?

yes it's a good idea. Can you reopen the MR so that we can run Appveyor tests?

JonathanSalwan:
> not sure if we need to go excessive and add 3.8 + 3.9 + 3.10 for win32 too instead of just x64?

yes it's a good idea. Can you reopen the MR so that we can run Appveyor tests?

PR #1144, opened 2022-06-27
[WIP] Add support for building Python Wheel packages

cnheitman:
This is the first step to provide users the chance to install Triton using `pip` (`pip install triton-library`).

This PR adds a `setup.py` script with support for Linux and Windows. Now you can build and install Triton's Python package by running `pip install .` from Triton's main folder. This will invoke `cmake`, `cmake --build` and then install the generated package.

It is possible to set custom builds for `Z3`, `Bitwuzla` and `Capstone` using environment variables (for example, `Z3_INCLUDE_DIRS` and `Z3_LIBRARIES}`). I tried this on Ubuntu and Windows (for Python 3.8) and it seems to be working fine.

You can also generate the `.whl` package in Linux by running:

```bash
# Install build package
pip install build
# Set environment variables for custom build (if you want/need)
export Z3_LIBRARIES=<path/to/z3/lib>
export Z3_INCLUDE_DIRS=<path/to/z3/include>
# ...
# Build the .whl package
python -m build --wheel
```

This PR also adds a workflow for building Triton's Python packages for both Linux and Windows (currently, it generates two artifacts: `triton_library-1.0.0-cp38-cp38-linux_x86_64.whl` and `triton_library-1.0.0-cp39-cp39-win_amd64.whl`). The goal is to build the Python packages with `Z3` and `Bitwuzla` linked statically for both Linux and Windows, and to use `wide-integer` instead of `Boost`, in order to make them as standalone as possible.

Pending tasks:

1. Build Bitwuzla in the workflow (WIP)
2. Link Visual C++ Runtime statically
3. Use VCPKG to build the packages
4. Add support for macOS
5. Tweaks here and there (for instance, provide builds for more Python versions)

Any help on the following will be much appreciated! :

* Item 2. This shouldn't be hard but as I'm not very familiar with Windows development I don't know which is the right way to do it.
* Item 3. Ideally, we should use something similar to the `cmake.yml` workflow. As I'm not familiar with VCPKG I went for the "standard" approach. However, the VCPKG seems like a better way to do it.
* Item 4. I don't have a working environment to test this but I don't think it will differ much from the Linux one.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -1,0 +1,121 @@
+name: Build Python Package
+
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  build-nix:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Upgrade pip version
+        run: |
+          python -m pip install -U pip
+
+      - name: Install build package
+        run: |
+          python -m pip install build
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get install python-setuptools
+
+      - name: Download Z3
+        run: |
+          wget https://github.com/Z3Prover/z3/releases/download/z3-4.8.17/z3-4.8.17-x64-glibc-2.31.zip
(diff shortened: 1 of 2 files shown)


JonathanSalwan:
This is an awesome job Chris! I will give it a try as soon as possible and see how can I help you on points that you mentioned.

JonathanSalwan:
I've pushed a new branch `dev-pip-intsall` and merged the MR on this branch so that we can work on the branch directly and test the Github CI. I also edited your first comment to add check boxes on pending tasks. I've also added one task (LLVM support).

Issue #1145, opened 2022-06-29
Add support for building Python Wheel packages


JonathanSalwan:
This refers to #1144.

### Pending tasks

- [ ] LLVM support (WIP: Linux and MacOS ok)
- [ ] Bitwuzla support (WIP: Linux and MacOS ok)
- [x] ~~Artifacts on Linux with all Python versions~~
- [x] ~~Artifacts on Windows with all Python versions~~
- [x] ~~Artifacts on macOS with all Python versions~~
- [x] ~~Supporting Triton's Interfaces~~

Any help on the following will be much appreciated! :

### Building the `.whl` by default

```console
$ python -m build --wheel
* Creating venv isolated environment...
* Installing packages in isolated environment... (setuptools >= 40.8.0, wheel)
* Getting dependencies for wheel...
running egg_info
creating triton_library.egg-info
[...]
* Installing packages in isolated environment... (wheel)
* Building wheel...
running bdist_wheel
running build
running build_ext
-- The C compiler identification is GNU 11.3.0
-- The CXX compiler identification is GNU 11.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
[...]
adding 'triton_library-1.0.0.dist-info/WHEEL'
adding 'triton_library-1.0.0.dist-info/top_level.txt'
adding 'triton_library-1.0.0.dist-info/RECORD'
removing build/bdist.linux-x86_64/wheel
Successfully built triton_library-1.0.0-cp39-cp39-linux_x86_64.whl
```

### Building the `.whl` with a specific interface

```console
$ LLVM_INTERFACE=On python -m build --wheel
```

By default, interfaces are defined as:

 - Z3_INTERFACE: `On`
 - LLVM_INTERFACE: `Off`
 - BITWUZLA_INTERFACE: `Off`
 - BOOST_INTERFACE: `Off`

### Installing the package with pip

```console
$ pip install . --user
Processing /home/jonathan/Works/Tools/Triton
  Preparing metadata (setup.py) ... done
Building wheels for collected packages: triton-library
  Building wheel for triton-library (setup.py) ... done
  Created wheel for triton-library: filename=triton_library-1.0.0-cp39-cp39-linux_x86_64.whl size=1518902 sha256=aeb378b42f39a7bff36bc2037aca1fd3fd94d37227609ebd719d6f7afde95ccb
[...]

PR #1155, opened 2022-07-08
Change wheel package version number

cnheitman:
This PR sets the version number for the Wheel package and adjusts related scripts and workflow (issue #1145).


JonathanSalwan:
@cnheitman, oh I forgot: can you add comments in src/scripts/docker/* files so that we know for what they are used? Very impressive job though!

cnheitman:
Thanks, @JonathanSalwan o/ 

I just pushed a new commit which adds the comments.

Issue #1159, opened 2022-07-13
divide-by-zero when using div with symbolic registers 


kettenbruch:
Since divide instructions like [`div_s`](https://github.com/JonathanSalwan/Triton/blob/master/src/libtriton/arch/x86/x86Semantics.cpp#L5440) evaluate their divisor,
symbolic registers trigger a Divide by zero exception. The emulator loops on the div instruction forever. 
The zero checking was added in [this](https://github.com/JonathanSalwan/Triton/commit/7fdc0ed929d768d8bd7be2449adcb5da54032048) commit.

minimal working example:
```
#!/usr/bin/env python
import sys
from triton import *

CODE = b"\x49\xC7\xC4\x01\x00\x00\x00\x48\xC7\xC0\x01\x00\x00\x00\x48\xC7\xC2\x01\x00\x00\x00\x49\xF7\xF4\xC3"
'''
0x1000 | 49C7C4 01000000         | mov r12,1
0x1007 | 48C7C0 01000000         | mov rax,1 
0x100e | 48C7C2 01000000         | mov rdx,1
0x1015 | 49F7F4                  | div r12  (jump here)
0x1018 | C3                      | ret      
'''
def emulate(ctx, pc):
    while pc:
        opcode = ctx.getConcreteMemoryAreaValue(pc, 16)
        instruction = Instruction(pc, opcode)
        if ctx.processing(instruction) != EXCEPTION.NO_FAULT:
            print('Error processing %s\n' % (str(instruction)))
            break

        print(instruction)
        pc = ctx.getRegisterAst(ctx.registers.rip).evaluate()
    return

def main():
    ctx = TritonContext(ARCH.X86_64)
    ast = ctx.getAstContext()

    # declare inputs
    ctx.symbolizeRegister(ctx.registers.rax, "rax")
    ctx.symbolizeRegister(ctx.registers.rdx, "rdx")
    ctx.symbolizeRegister(ctx.registers.r12, "r12")
    ctx.setConcreteMemoryAreaValue(0x1000, CODE)
    emulate(ctx, 0x1015)

    raxast = ast.unroll(ctx.getSymbolicRegister(ctx.registers.rax).getAst())
    rdxast = ast.unroll(ctx.getSymbolicRegister(ctx.registers.rdx).getAst())
    print(f"rax = %s\nrdx = %s\n" % (raxast, rdxast))

if __name__ == '__main__':
    sys.exit(main())

''' expected:
0x1015: div r12
[...]

PR #1160, opened 2022-07-13
Exclude symbolic results from exceptions

kettenbruch:
added test cases for  #1159 

Now the results are only evaluated when they are not symbolic.

### DIFF
--- a/src/libtriton/arch/x86/x86Semantics.cpp
+++ b/src/libtriton/arch/x86/x86Semantics.cpp
@@ -5459,7 +5459,7 @@ namespace triton {
             /* Apply the taint */
             expr->isTainted = this->taintEngine->taintUnion(ax, src);
             /* Divide error */
-            if (result->evaluate() > 0xff) {
+            if (!result->isSymbolized() && result->evaluate() > 0xff) {
               this->exception = triton::arch::FAULT_DE;
               return;
             }
@@ -5485,7 +5485,7 @@ namespace triton {
             /* Apply the taint for DX */
             expr2->isTainted = this->taintEngine->taintUnion(dx, src);
             /* Divide error */
-            if (temp->evaluate() > 0xffff) {
+            if (!temp->isSymbolized() && temp->evaluate() > 0xffff) {
               this->exception = triton::arch::FAULT_DE;
               return;
             }
@@ -5511,7 +5511,7 @@ namespace triton {
             /* Apply the taint for EDX */
             expr2->isTainted = this->taintEngine->taintUnion(edx, src);
             /* Divide error */
-            if (temp->evaluate() > 0xffffffff) {
+            if (!temp->isSymbolized() && temp->evaluate() > 0xffffffff) {
               this->exception = triton::arch::FAULT_DE;
               return;
             }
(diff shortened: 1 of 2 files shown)


JonathanSalwan:
Thanks for this MR!

Yesterday I've reverted some `idiv` checks (b0d93a5d3d42ada59b04e56ba29c769418fba3d1) because they were wrong. Divide errors are raised for [different reasons](http://shell-storm.org/x86doc/IDIV.html). I think this is  why unit tests failed.

> Now the results are only evaluated when they are not symbolic.

I'm not sure about this =/. For example what if your node is symbolic but the result of `temp` does not fit in the dst register (`>0xffffffffffffffff`), do you continue the execution while it should stop? I think it leads to an incorrect path predicate. What do you think?

kettenbruch:
> I think it leads to an incorrect path predicate
True, before there were any exceptions in `div_s` Triton would just `(_ extract 63 0)` the lower 64 bits and continue execution while a real machine would throw a `#DE`.

Still, we should be able to use `div` on purely symbolic registers (see [test_3](https://github.com/JonathanSalwan/Triton/pull/1160/commits/930621d424bdc3a5666dbb926a8612305989863c#diff-2d3303c526cc9f432b4e712645a36dde7df757e20eafc65c15e60383c9bbc896R590))
The problem is that we evaluate the symbolic register to 0, divide by it and get `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF` which is too big to fit a 64bit register. 
The concrete value of temp is too big, but this doesn't matter in test_3 as everything is symbolic. If we throw an exception, Triton gets stuck at evaluating the `div` instruction and we never get a complete AST. 

I got a CTF which ran fine with some [tweaking](https://gist.github.com/kettenbruch/3c21b58c47d104c608c3090c62adc27a#file-solver-py-L98) in Triton v0.9 but got stuck on the `div` instruction in dev-v1.0 [here](https://gist.github.com/kettenbruch/3c21b58c47d104c608c3090c62adc27a).

JonathanSalwan:
> If we throw an exception, Triton gets stuck at evaluating the div instruction

Yes indeed this one is another error. We have to reset the state as `NO_FAULT` for every new instructions executed. 1960dcb3a3af3fa7feb68ddd692b57a4f3247622 should fixes this.

JonathanSalwan:
Regarding concret vs symbolic, i'm still convinced that we have to define the state as `DE` even is symbolic or not. Now that the state is not stuck in `DE` mode with the previous patch (1960dcb3a3af3fa7feb68ddd692b57a4f3247622), you have to choice to continue or not the execution. For example with your CTF challenge your code may looks like this:

```python
def emulate(ctx, pc):
    while pc:
        opcode = ctx.getConcreteMemoryAreaValue(pc, 16)
        instruction = Instruction(pc, opcode)
        exc = ctx.processing(instruction)

        if instruction.isSymbolized():
            print("\033[92m" + str(instruction) + "\033[0m")
        else:
            print(instruction)

        # still need to constrain RDX to avoid #DE on a real machine (RDX:RAX too big)
        # otherwise we get incorrect results (in v0.9 and dev-v1.0)
        if pc == 0x1090:  # div instruction address
            ctx.pushPathConstraint(ctx.getRegisterAst(ctx.registers.rdx) < 0x1000)

        if exc != EXCEPTION.NO_FAULT:
            print(f'Error processing {str(instruction)}: {exc:d}\n')
            pc = instruction.getNextAddress() # <- skip the DE exception
        else:
            pc = ctx.getRegisterAst(ctx.registers.rip).evaluate()
    return
```

kettenbruch:
I see. I'll handle the exception in user code as you suggested.
Thanks for the code and huge thanks for Triton in general, it's a really cool project !

PR #1172, opened 2022-08-10
AARCH64: Fix instruction memory access size when it's specified by instruction (ldrb, etc)

Antwy:
When disassembling load/store instructions memory access size is set according to the destination register size. This is wrong for some instructions like ldrb w8, [sp, #0xc] as it takes only 1 byte.


SweetVishnya (approved):
Yeah, this merge definitely fixes memory access sizes for AARCH64. We need proper memory access sizes to handle switch tables and symbolic pointers.

JonathanSalwan:
Hey @Antwy,

Thx for this MR. Can you apply the same patch for arm32? I think we may have the same issue.

PR #1175, opened 2022-08-31
Add vcpkg installation instructions

Cheney-W:
Triton is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for triton and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build triton, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here](https://github.com/microsoft/vcpkg/blob/master/ports/triton/portfile.cmake) is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,21 @@ Note that if you use AppVeyor's artefacts, you probably have to install the [Vis
 packages for Visual Studio 2012.
 
 
+### Installing from vcpkg
+
+The triton port in vcpkg is kept up to date by Microsoft team members and community contributors. The url of vcpkg is: https://github.com/Microsoft/vcpkg . You can download and install triton using the vcpkg dependency manager:
+
+```shell
+git clone https://github.com/Microsoft/vcpkg.git
+cd vcpkg
+./bootstrap-vcpkg.sh  # ./bootstrap-vcpkg.bat for Windows
+./vcpkg integrate install
+./vcpkg install triton
+```
+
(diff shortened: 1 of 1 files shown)


JonathanSalwan:
Hey @Cheney-W,

Thanks to have taken time to update our Readme file. I'm not familiar with vcpkg and I would like to know if those files [1](https://github.com/JonathanSalwan/Triton/blob/master/vcpkg.json), [2](https://github.com/JonathanSalwan/Triton/blob/master/CMakePresets.json) and [3](https://github.com/JonathanSalwan/Triton/tree/master/vcpkg/triplets) are still mandatory to install Triton via your packet manager?

Cheney-W:
1 and 3 yes, 2 no.

1 is a custom file for the installation package in vcpkg manifest mode, which can be used to limit the version of the package to be installed. It is equivalent to a configuration file for a specific package.

3 are some custom triplet files, which are different from our official triplet files. They are generally added by users to meet their own specific config or link options. They can be used when installing packages to use Install packages under certain conditions, for example, build only in release mode, not in debug mode.

2 has nothing to do with vcpkg, just configure some information of vcpkg into itself, but if you don't actively find the third-party library installed through vcpkg in the cmakelist file, you can assume that information does not exist.

PR #1176, opened 2022-09-01
AARCH64: add pushing path constraint in TBNZ, TBZ semantics

Antwy:
Include TBNZ, TBZ branch instructions into path predicate by pushing path constraints


src/libtriton/arch/arm/aarch64/aarch64Semantics.cpp
```diff
@@ -3586,6 +3586,13 @@ namespace triton {
 
           /* Spread taint */
           expr->isTainted = this->taintEngine->setTaint(dst, this->taintEngine->isTainted(src1) | this->taintEngine->isTainted(src2));
+
+          /* Set condition flag */
+          if (this->astCtxt->bvlshr(op1, op2)->evaluate() != 0)
```
JonathanSalwan, line 3591: I think it's more something like this:
`this->astCtxt->extract(0, 0, this->astCtxt->bvlshr(op1, op2))->evaluate()` ?

PR #1188, opened 2022-09-29
Add x86 PSRLD, PSRLQ and PSRLW semantics

hexpell:


JonathanSalwan:
I'm fixing the Bitwuzla CI first and then I will review/merge your MR :)

JonathanSalwan:
Also, can you add some tests in this [file](https://github.com/JonathanSalwan/Triton/blob/dev-v1.0/src/testers/x86/unicorn_test_x86.py)? You just have to append to the `CODE` list some `psrlX` instructions with random values into registers. Try to combine a lot of combination like different sizes of register and memory.

PR #1200, opened 2022-11-03
Store instruction address in symbolic expression created from instructions.

ek0:
Hey @JonathanSalwan,

This is a small proposal, and hopefully a small enough change. If you could give some feedback it would be much appreciated :)

Currently, apart the disassembly string, symbolic expressions don't have any information regarding the original instruction that was used for lifting. Having the original address embedded in the symbolic expression would be useful to do backtracking using the backward slicing functionality (through `sliceExpressions`)

For example: looking at the dead store elimination simplification, it seems there's a [small hack](https://github.com/JonathanSalwan/Triton/blob/2b655f20528065cf70e0fa95e2d01b34a8ef6a17/src/libtriton/engines/symbolic/symbolicSimplification.cpp#L263) parsing the disassembly string to retrieve the original instruction address.

If we keep this information in the symbolic expression this would not be required.

- I've mainly added an `address` field to the `SymbolicExpression` object. With a setter and a getter. And everytime a new expression is created from an instruction we set this value.
- If address is unknown, it is set to 0. Maybe we could also use a `std::optional` instead of a `uint64_t` to reflect the absence of address?
- Maybe I could just put the `setAddress(getAddress())` in `Instruction::addSymbolicExpression`. But from an ownership point of view it feels better if it is set as it is done in this PR.

### DIFF
--- a/src/libtriton/engines/symbolic/symbolicEngine.cpp
+++ b/src/libtriton/engines/symbolic/symbolicEngine.cpp
@@ -973,6 +973,7 @@ namespace triton {
         if (this->isAlignedMode() && this->isArrayMode() == false) {
           const SharedSymbolicExpression& aligned = this->newSymbolicExpression(node, MEMORY_EXPRESSION, "Aligned optimization - " + comment);
           aligned->setOriginMemory(mem);
+          aligned->setAddress(inst.getAddress());
           this->addAlignedMemory(address, writeSize, aligned);
           /* Refresh the current id to not link the aligned expression to the instruction */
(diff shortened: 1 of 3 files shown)


JonathanSalwan:
Hey @ek0,

Thanks for this MR. I do remember thinking about that feature when implementing the **small hack**, but I do not remember why I did not do it... (maybe to lazy?). Let me few days to see if there is no hidden issues behind this feature and if everything looks good to me, I will merge this one :).

Cheers,

ek0:
Sure thing, let me do this quickly

PR #1215, opened 2022-12-23
Build wheel for Python 3.11

Andrix44:
Tested the Github actions artifact out on my machine and it installed successfully.
Most of the build time is spent on the dependencies so this shouldn't affect it by much.


cnheitman (approved):
Looks good, we can merge.

PR #1229, opened 2023-02-08
Switch to CMake's `Python3` package

mrexodia:
The `find_package(PythonInterp)` and `find_package(PythonLibs)` are deprecated and on my system they were returning results from two different python versions installed (interpreter from my PATH and libraries from homebrew's python). This caused me to run into this (very confusing) error: https://github.com/JonathanSalwan/Triton/issues/1145#issuecomment-1206549486

The `Python3` package fixed this behaviour and I made sure to keep the interface compatible (`PYTHON_XXX` variables work the same). The default behaviour (you just run `cmake -B build`) now also finds the python the user would expect (the one that's currently in the PATH) and things are working properly from a virtual environment. Here is a `requirements.txt` to test:

```
git+https://github.com/mrexodia/Triton@8ee04d36ff732f0a89dfe7609806ad45addc0e45#egg=triton-library
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,18 +46,56 @@ add_custom_target(check
 
 if(PYTHON_BINDINGS)
     message(STATUS "Compiling with Python bindings")
+
+    if(NOT PYTHON_VERSION AND NOT PYTHON_EXECUTABLE)
+        # Find the python version the user has in the PATH
+        # This prevents an issue where an unexpected python version is used
+        # (eg the system/homebrew python in a virtual environment)
+        find_program(PYTHON_EXECUTABLE
+            NAMES python3 python
+            NO_PACKAGE_ROOT_PATH
+            NO_CMAKE_PATH
+            NO_CMAKE_ENVIRONMENT_PATH
+            NO_CMAKE_SYSTEM_PATH
+            NO_CMAKE_INSTALL_PREFIX
+            REQUIRED
+        )
+    endif()
+
+    if(PYTHON_EXECUTABLE)
+        # Make sure the specified python is not fully broken
+        execute_process(COMMAND ${PYTHON_EXECUTABLE} --version
+            RESULT_VARIABLE PYTHON_RESULT
+            OUTPUT_QUIET
+        )
+        if(NOT PYTHON_RESULT EQUAL 0)
+            message(FATAL_ERROR "Broken python: ${PYTHON_EXECUTABLE}")
+        endif()
+    endif()
+
+    # Set the hints for the Python3 package
+    # Reference: https://cmake.org/cmake/help/latest/module/FindPython3.html#hints
+    # The PYTHON_XXX variables are not used by FindPython3, but they are translated
+    # for backwards compatibility with previous Triton versions
+    set(Python3_EXECUTABLE "${PYTHON_EXECUTABLE}")
+    if(PYTHON_LIBRARY)
+        set(Python3_LIBRARY "${PYTHON_LIBRARY}")
+    elseif(PYTHON_LIBRARIES)
+        set(Python3_LIBRARY "${PYTHON_LIBRARIES}")
+    endif()
+    if(PYTHON_INCLUDE_DIRS)
+        set(Python3_INCLUDE_DIR "${PYTHON_INCLUDE_DIRS}")
+    endif()
+    set(Python3_FIND_VIRTUALENV "FIRST")
+
+    find_package(Python3 ${PYTHON_VERSION} COMPONENTS Interpreter Development REQUIRED)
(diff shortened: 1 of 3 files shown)


src/libtriton/CMakeLists.txt
```diff
@@ -272,14 +272,17 @@ target_include_directories(triton BEFORE PUBLIC
...
-    ${PYTHON_LIBRARIES}
     ${Boost_LIBRARIES}
     ${Z3_LIBRARIES}
     ${LLVM_LIBRARIES}
     ${BITWUZLA_LIBRARIES}
     ${CAPSTONE_LIBRARIES}
 )
 
+if(PYTHON_BINDINGS)
+    target_link_libraries(triton PRIVATE Python3::Module)
```
JonathanSalwan, line 283: According to unit tests, i'm not sure it should be private.

CMakeLists.txt
```diff
@@ -46,18 +46,56 @@ add_custom_target(check
...
+        set(Python3_INCLUDE_DIR "${PYTHON_INCLUDE_DIRS}")
+    endif()
+    set(Python3_FIND_VIRTUALENV "FIRST")
+
+    find_package(Python3 ${PYTHON_VERSION} COMPONENTS Interpreter Development REQUIRED)
+    message(STATUS "Python3 includes: ${Python3_INCLUDE_DIRS}")
+    message(STATUS "Python3 libraries: ${Python3_LIBRARIES}")
+
+    # For backwards compatibility with existing CMake
+    set(PYTHON_EXECUTABLE "${Python3_EXECUTABLE}")
```
JonathanSalwan, line 96: Why not just using PYTHON_EXECUTABLE? I think we dropped the Python2.7 support.

CMakeLists.txt
```diff
@@ -46,18 +46,56 @@ add_custom_target(check
...
+        set(Python3_LIBRARY "${PYTHON_LIBRARY}")
+    elseif(PYTHON_LIBRARIES)
+        set(Python3_LIBRARY "${PYTHON_LIBRARIES}")
     endif()
-    include_directories(${PYTHON_INCLUDE_DIRS})
-    add_definitions("-DPYTHON_LIBRARIES=\"${PYTHON_LIBRARIES}\"")
+    if(PYTHON_INCLUDE_DIRS)
+        set(Python3_INCLUDE_DIR "${PYTHON_INCLUDE_DIRS}")
+    endif()
+    set(Python3_FIND_VIRTUALENV "FIRST")
```
JonathanSalwan, line 89: For what `FIRST` is used? Is it a standard?

mrexodia: It’s documented in the link above. It’s so the python from the virtualenv is picked first (and then the system one). This is desirable because then you can use Triton from requirements.txt/pyproject.toml

CMakeLists.txt
mrexodia: That variable is not set anymore. The new variable is named `Python3_EXECUTABLE`.

src/libtriton/CMakeLists.txt
mrexodia: Yeah I’ll fix this next week. I think there is an issue with the way the unit tests are set up. The triton shared library (pyd python module) should just import from python. The unit tests shouldn’t have to know about python at all?

mrexodia: Maybe to clarify a bit: a PUBLIC dependency would become a part of the Triton CMake package. This seems incorrect conceptually, because a package consumer does not need to know whether the python bindings are included in the shared library or not (and thus doesn’t have to find Python, etc)

cnheitman:
Hi @mrexodia, what's the status of this PR? Can I attempt to rebase and merge it?

src/libtriton/CMakeLists.txt
cnheitman: On my side, I had to add `target_link_libraries(triton PRIVATE ${Python3_LIBRARIES})` in additon to the existing one inside the `PYTHON_BINDINGS` if block:
```cmake
if(PYTHON_BINDINGS)
    target_link_libraries(triton PRIVATE ${Python3_LIBRARIES})
    target_link_libraries(triton PRIVATE Python3::Module)
endif()
```
Otherwise I got `undefine reference` errors when it tries to link the python library. Any idea why @mrexodia ?

mrexodia: In theory this is not necessary, but welcome to CMake/Python land where nothing makes sense 🤷‍♂️

cnheitman: Ok, I'll add that line, rebase and merge then. Everything else seems to be working fine.

PR #1236, opened 2023-03-28
Expose CPU memory from cpuInterface.

archercreat:
Currently there's no way to const copy cpu memory without invoking copy constructor. This pull request adds `getAllMemory` api which returns const reference to an existing std::unordered_map.

### DIFF
--- a/src/libtriton/arch/architecture.cpp
+++ b/src/libtriton/arch/architecture.cpp
@@ -172,6 +172,12 @@ namespace triton {
       return this->cpu->getAllRegisters();
     }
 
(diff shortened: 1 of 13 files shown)


archercreat:
> Maybe, we can use better naming like getConcreteMemory?

I called it because there is already `getAllRegisters` function which does the same but i'm ok to change.

JonathanSalwan:
Hi @archercreat, thanks for this feature! `getAllRegisters` is used to returns registers of the CPU. As your feature returns the memory concrete state, I think that `getConcreteMemory` would be appropriate as @SweetVishnya said :)

SweetVishnya (approved):

PR #1263, opened 2023-07-14
Upgrade Bitwuzla to v0.1.0

cnheitman:
This PR upgrades Bitwuzla code in Triton to work with Bitwuzla v0.1.0.


JonathanSalwan:
The previous merge of #1229 killed the Appveyor CI. Can we fix it in this MR too so that we can confirm this change also works on Windows?

mrexodia:
As a sidenote, GitHub actions supports Windows runners so it might be worth investing some time to move there. Especially considering AppVeyor became extremely slow…

JonathanSalwan (approved):

PR #1273, opened 2023-08-17
Clang format

ZehMatt:
I've tried to minimize the diff but also didn't want to exactly match it to get somewhere in the middle of readability and consistency. I also enabled the option to auto comment the end of the namespace, this is a bit inconsistent at the moment and we should either turn that option off or remove the ones that don't fit, I personally think its best to let clang-format do the commenting so it won't be missed and it will be consistent across the board.

Probably need a few more files I should look into, clang-format can get quite notorious about tables.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,29 @@
+---
+BasedOnStyle: Webkit
+IndentWidth: 2
+Standard: Cpp11
+UseTab: Never
+TabWidth: 2
+MaxEmptyLinesToKeep: 1
+ColumnLimit: 120
+BreakBeforeBraces: Attach
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: Never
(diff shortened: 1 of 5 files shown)


SweetVishnya:
We could also add clang-format checks to CI

ZehMatt:
> We could also add clang-format checks to CI

I could do that but its best to have all of the files formatted first

JonathanSalwan:
Thanks for this MR,

> I could do that but its best to have all of the files formatted first

Before updating all files, we should first agree on a format :). I will play with the example you provided and  get back to you as soon as I can.

ZehMatt:
> Before updating all files, we should first agree on a format :). 

Definitely, that is also what I meant.

> I will play with the example you provided and get back to you as soon as I can.

No rush here, thanks for looking into it.

ZehMatt:
> This is a nice addition :)
> 
> @ZehMatt Could you modify the PR so it merges into `dev-v1.0` instead of `master`?

Not sure why GitHub decided to pick master, I branched off dev-1.0 🤷‍♂️  , changed it.

cnheitman:
Thanks, @ZehMatt !

Issue #1248, opened 2023-05-14
Properly Installing in windows, under conda


SirWaffle:
I am unsure how to properly install triton in windows.
What I've done:
- cloned the repo
- built triton using vcpkg

now im left unsure what to do. I saw on a previous issue, that the built DLL needs to be renamed to *.pyd and placed in the python path, which I have done. However, I am still unable to import and use triton.

I tried to pip install . inside of the triton checkout, but package install fails due to the wheel failing.

JonathanSalwan:
Also check that all deps like z3 are into the same folder than triton.pyd

SirWaffle:
I copied and pasted this set of files in a few places....
the capstone DLL
the lib Z3 dll
the triton DLL
i also duplicated all of them, and changed the extension to pyd

I get this error when trying to import triton:
Python.Runtime.PythonException: 'ImportError : dynamic module does not define module export function (PyInit_triton)'

any ideas?

JonathanSalwan:
`PyInit_triton` makes me thing about an issue around Python version (compile vs runtime). Is the Python version used during the compile is the same than the one used at runtime?

PavelKotov1:
Hello, dear friends!
Pretty much the same situation installing triton and referencing it from python script.
Could somebody in the name of charity and goodwill explain how to handle this problem? Detailed instruction would be very much appreciated since i am out of options after trying all recommended steps...
Thanks in advance!

JonathanSalwan:
Sorry guys, I can't help you on that side as i do not have a Windows on hand.

Andrix44:
The last time I built it for Windows, I downloaded the source and all the dependencies (LLVM, Z3, Capstone...) then I opened the folder with cmake-gui. You have to manually enable features and enter the paths to your downloaded libraries. You should also enable PYTHON_BINDINGS_AUTOCOMPLETE. Click configure until there are no errors then click generate and make sure you only have warnings and no errors. Then you can open the solution in VS. Build python-triton and triton_autocomplete. Now you should have the .pyd and .pyi files that you can copy to your python site-packages directory. If you get any errors while building you probably didn't set up the dependecies correctly in CMAKE.

PavelKotov1:
Cher Andrix44! 
Well, that sounds like a really good answer! 
I shall definitely try it ASAP and then report back the results to the community!
Thanks a lot!

trustednstaller:
[...]

PR #1283, opened 2023-09-04
Fixed the way in which windows is build

cctv130:


CMakeLists.txt
```diff
@@ -19,18 +25,30 @@ set(VERSION_BUILD ${BUILD_NUMBER})
...
 include(CMakeDependentOption)
 
+
+
 option(ASAN                              "Enable the ASAN linking"                         OFF)
 option(BITWUZLA_INTERFACE                "Use Bitwuzla as SMT solver"                      OFF)
 option(BUILD_SHARED_LIBS                 "Build a shared library"                          ON)
 option(GCOV                              "Enable code coverage"                            OFF)
-option(LLVM_INTERFACE                    "Use LLVM for lifting"                            OFF)
+option(LLVM_INTERFACE                    "Use LLVM for lifting"                            ON)
```
JonathanSalwan, line 45: We should not compile LLVM by default. Must be an user option.

CMakeLists.txt
```diff
@@ -19,18 +25,30 @@ set(VERSION_BUILD ${BUILD_NUMBER})
...
 option(MSVC_STATIC                       "Use statically-linked runtime library"           OFF)
 option(Z3_INTERFACE                      "Use Z3 as SMT solver"                            ON)
-option(BOOST_INTERFACE                   "Use Boost as multiprecision library"             ON)
-option(PYTHON_BINDINGS_AUTOCOMPLETE      "Generate an autocomplete stub file"              OFF)
+option(BOOST_INTERFACE                   "Use Boost as multiprecision library"             OFF)
```
JonathanSalwan, line 48: Ok for switching boost ON to OFF by default.

CMakeLists.txt
```diff
@@ -19,18 +25,30 @@ set(VERSION_BUILD ${BUILD_NUMBER})
...
+option(PYTHON_BINDINGS_AUTOCOMPLETE      "Generate an autocomplete stub file"              ON)
```
JonathanSalwan, line 49: We should not add `PYTHON_BINDINGS_AUTOCOMPLETE` by default.

CMakeLists.txt
```diff
@@ -7,9 +7,15 @@
...
 cmake_minimum_required(VERSION 3.20)
+set(CMAKE_TRY_COMPILE_TARGET_TYPE "STATIC_LIBRARY")
 
 set(TRITON_ROOT "${CMAKE_CURRENT_LIST_DIR}")
 
+#C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/Llvm
+
+
+
+
```
JonathanSalwan, line 18: Maybe we can remove this?

CMakeLists.txt
```diff
@@ -19,18 +25,30 @@ set(VERSION_BUILD ${BUILD_NUMBER})
 # Used for write_basic_package_version_file()
 project(triton VERSION ${VERSION_MAJOR}.${VERSION_MINOR})
 
+
+
+if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_GENERATOR_TOOLSET STREQUAL "ClangCL")
+    SET(CMAKE_CXX_FLAGS "/EHsc")
```
JonathanSalwan, line 31: If you can add a comment about what those flags means :)

JonathanSalwan:
Just to fully understand, what is the problem of compile Triton on windows? Does this [step](https://github.com/jonathansalwan/Triton#windows) not works?

cctv130:
> Just to fully understand, what is the problem of compile Triton on windows? Does this [step](https://github.com/jonathansalwan/Triton#windows) not works?

This piece is modified according to #1248

CMakeLists.txt
cctv130: > If you can add a comment about what those flags means :)

If you're using LLVM on windows, clang-cl is recommended over MSVC.

cctv130: > If you can add a comment about what those flags means :)

If you want to use the clang - cl, requires this setting EHsc, prevent errors

cctv130:
> Just to fully understand, what is the problem of compile Triton on windows? Does this [step](https://github.com/jonathansalwan/Triton#windows) not works?

For building projects on windows, I have full instructions on llvm builds and a full triton installation at WindowsBuild.txt.

cctv130:
> Just to fully understand, what is the problem of compile Triton on windows? Does this [step](https://github.com/jonathansalwan/Triton#windows) not works?

I am very sorry that due to my network problem, the local docker cannot access the external network, so I can only use github ci to compile, one compilation failed, and the others succeeded. At present, there is another problem, docker compilation can not find the path of meson, I do not know how to modify this path in docker. Try to change the DOCKERFILE.
Build Python Package ：the problem is that meson I see is already installed, there is a warning that meson is not added to the path.

cctv130:
> Just to fully understand, what is the problem of compile Triton on windows? Does this [step](https://github.com/jonathansalwan/Triton#windows) not works?

All builds completed successfully, please check.

.idea/Triton.iml
```diff
@@ -0,0 +1,12 @@
...
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$" />
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+  <component name="PyDocumentationSettings">
+    <option name="format" value="PLAIN" />
+    <option name="myDocStringFormat" value="Plain" />
+  </component>
+</module>
```
JonathanSalwan, line 12: What are all files in the `.idea` directory used for?

CMakeLists.txt
```diff
@@ -19,9 +19,19 @@ set(VERSION_BUILD ${BUILD_NUMBER})
...
+        set(CMAKE_TRY_COMPILE_TARGET_TYPE "STATIC_LIBRARY")
+        set(CMAKE_CXX_FLAGS "/EHsc")
+    endif()
+endif ()
+
+
 # Define cmake options
 include(CMakeDependentOption)
 
+
```
JonathanSalwan, line 34: You can remove this line :)

CMakeLists.txt
```diff
@@ -32,6 +42,10 @@ option(Z3_INTERFACE                      "Use Z3 as SMT solver"
 option(BOOST_INTERFACE                   "Use Boost as multiprecision library"             ON)
 option(PYTHON_BINDINGS_AUTOCOMPLETE      "Generate an autocomplete stub file"              OFF)
 
+
+
+
+
```
JonathanSalwan, line 48: You can remove these lines :)

setup.py
```diff
@@ -4,6 +4,7 @@
 ##  This program is under the terms of the Apache License 2.0.
 ##
 
+
```
JonathanSalwan, line 7: You can remove this line :)

src/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 add_subdirectory(libtriton)
-
+
```
JonathanSalwan, line 2: You can remove this line :)

src/scripts/docker/build-wheel-linux.sh
```diff
@@ -14,6 +14,9 @@
 set -e
 # set -x  # Debugging
 
+
+
+
```
JonathanSalwan, line 19: You can remove these lines :)

src/scripts/docker/build-wheel-linux.sh
```diff
@@ -30,42 +33,69 @@ export BITWUZLA_LIBRARIES=$DEPENDENCIES_DIR/bitwuzla/install/lib64/libbitwuzla.s
...
-$PYTHON_BINARY -m build --wheel --outdir $WHEEL_DIR/linux_x86_64
+#export PYTHON_BINARY=/opt/_internal/cpython-3.11.4/bin/python
+#export PYTHON_INCLUDE_DIRS=$($PYTHON_BINARY -c "from sysconfig import get_paths; print(get_paths()['include'])")
+#export PYTHON_LIBRARY=$($PYTHON_BINARY -c "from sysconfig import get_paths; print(get_paths()['include'])")
+conda activate py311
+python -m pip install build
+python --version
+python -m build --wheel --outdir $WHEEL_DIR/linux_x86_64
 
+python -m pip install auditwheel
```
JonathanSalwan, line 98: @cnheitman is this update ok for you? (as I know you worked a lot on this file too).

JonathanSalwan:
Thanks a lot @cctv130! Just some typo like removing empty lines and we are good. I just need the confirmation from @cnheitman regarding the `build-wheel-linux.sh` file and we are good :)

cnheitman:
@JonathanSalwan I'll take a look and test it on my local setup tomorrow.

cctv130:
> Thanks a lot @cctv130! Just some typo like removing empty lines and we are good. I just need the confirmation from @cnheitman regarding the `build-wheel-linux.sh` file and we are all good :)

Latest has been repair according to your request to delete the unnecessary files and space, according to my work yesterday, build - wheel - Linux. Sh, can not find a python version of the specified absolute path meson environment variables also can not find, then add the virtual environment, build python package will compile, maybe the image version is updated.

CMakeLists.txt
```diff
@@ -19,6 +19,15 @@ set(VERSION_BUILD ${BUILD_NUMBER})
 # Used for write_basic_package_version_file()
 project(triton VERSION ${VERSION_MAJOR}.${VERSION_MINOR})
 
+
+if(CMAKE_HOST_SYSTEM_NAME MATCHES "Windows")
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_GENERATOR_TOOLSET STREQUAL "ClangCL")
+        set(CMAKE_TRY_COMPILE_TARGET_TYPE "STATIC_LIBRARY")
+        set(CMAKE_CXX_FLAGS "/EHsc")
```
mrexodia, line 26: Please don't do this, most likely there is something wrong with your compilation environment. If you need custom flags, just pass them to cmake while configuring...

setup.py
```diff
@@ -90,10 +90,13 @@ def build_extension(self, ext):
 
         elif platform.system() == "Windows":
             cmake_args += [
-                '-G Visual Studio 17 2022',
+                '-G Visual Studio 17 2022',"-DCMAKE_CXX_STANDARD=20"
```
mrexodia, line 93: This should be part of the CMake file as a requirement for building Triton, why is it set here?

cctv130: > This should be part of the CMake file as a requirement for building Triton, why is it set here?

Please look at where you have any questions.

CMakeLists.txt
cctv130: > Please don't do this, most likely there is something wrong with your compilation environment. If you need custom flags, just pass them to cmake while configuring...

This is the content of the Windows in order to use the clang-cl to compile options, and there are some differences between MSVC, only in this way can we build triton clang-cl project success.
Passing it to cmake works, But if in cmakelist normal compilation,Will not affect other.

setup.py
cctv130: > This should be part of the CMake file as a requirement for building Triton, why is it set here?

The reason I put it here is that when I'm testing it, to be consistent with the llvm compiled version, this can be passed to the cmake configuration. Fixed.

cctv130: > This should be part of the CMake file as a requirement for building Triton, why is it set here?

Assume that if you are using c++ 17compile to llvm ,compile triton project also needs to c++ 17, example:using c++ 11 or c++ 20 other versions may be wrong

mrexodia:
I just tried building Triton `master` on Windows with ClangCL (latest Visual Studio update):

```
cmake -B build -T ClangCL "-DCMAKE_PREFIX_PATH:FILEPATH=d:\CodeBlocks\z3\install;d:\CodeBlocks\capstone\build\install;d:\CodeBlocks\llvm-12.0.0-rc1-install" -DLLVM_INTERFACE=ON
cmake --build build --config Release
```

It works without any modifications to the `CMakeLists.txt` (or passing the CXX_STANDARD_REQUIRED variable). I saw this issue before, it happens in one of the following circumstances:

1. Using a non-Microsoft version of `clang-cl` (eg downloaded from the latest LLVM releases, or modifying the VS installation)
2. Using CLion, where they made a major mistake and pass `-m64` to the clang command line, which completely breaks all builds. The issue was closed, but you can see it here: https://youtrack.jetbrains.com/issue/CPP-32063 (I am pretty confident it's not resolved yet, but perhaps it's only fixed in a prerelease).

To confirm everything works I also did a build with Ninja:

```
vs2022 x64 # open a VS2022 command line
cmake -B build "-DCMAKE_PREFIX_PATH:FILEPATH=d:/CodeBlocks/z3/install;d:/CodeBlocks/capstone/build/install;d:/CodeBlocks/llvm-12.0.0-rc1-install" -DLLVM_INTERFACE=ON "-DCMAKE_C_COMPILER:FILEPATH=c:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/Llvm/x64/bin/clang-cl.exe" "-DCMAKE_CXX_COMPILER:FILEPATH=c:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/Llvm/x64/bin/clang-cl.exe" -G Ninja -DCMAKE_BUILD_TYPE=Release
cmake --build
```

There are some pitfalls, notably you cannot pass `-DCMAKE_CXX_COMPILER=clang-cl.exe`, because this will trigger the usage of the `clang-cl.exe` in your LLVM installation (which causes errors are mentioned in point 1). Additionally forgetting to pass `-DCMAKE_BUILD_TYPE=Release` will error and you need to use `/` in the paths because of a CMake bug.

CMakeLists.txt
mrexodia: I wrote an elaborate comment here to continue the discussion: https://github.com/JonathanSalwan/Triton/pull/1283#issuecomment-1708045988

JonathanSalwan:
I agreed with that we should avoid as much as possible touching the CMake file and let the user provide his arguments according to his setup.

cctv130:
> 我刚刚尝试`master`使用 ClangCL（最新的 Visual Studio 更新）在 Windows 上构建 Triton：
> 
> ```
> cmake -B build -T ClangCL "-DCMAKE_PREFIX_PATH:FILEPATH=d:\CodeBlocks\z3\install;d:\CodeBlocks\capstone\build\install;d:\CodeBlocks\llvm-12.0.0-rc1-install" -DLLVM_INTERFACE=ON
> cmake --build build --config Release
> ```
> 
> 它无需任何修改`CMakeLists.txt`（或传递 CXX_STANDARD_REQUIRED 变量）即可工作。我以前见过这个问题，它发生在以下情况之一：
> 
> 1. 使用非 Microsoft 版本`clang-cl`（例如从最新的 LLVM 版本下载，或修改 VS 安装）
> 2. 使用 CLion，他们犯了一个重大错误并传递`-m64`到 clang 命令行，这完全破坏了所有构建。该问题已关闭，但您可以在这里看到它： https: [//youtrack.jetbrains.com/issue/CPP-32063](https://youtrack.jetbrains.com/issue/CPP-32063)（我非常有信心它尚未解决，但也许它只在预发布中修复）。
> 
> 为了确认一切正常，我还使用 Ninja 进行了构建：
> 
> ```
> vs2022 x64 # open a VS2022 command line
> cmake -B build "-DCMAKE_PREFIX_PATH:FILEPATH=d:/CodeBlocks/z3/install;d:/CodeBlocks/capstone/build/install;d:/CodeBlocks/llvm-12.0.0-rc1-install" -DLLVM_INTERFACE=ON "-DCMAKE_C_COMPILER:FILEPATH=c:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/Llvm/x64/bin/clang-cl.exe" "-DCMAKE_CXX_COMPILER:FILEPATH=c:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/Llvm/x64/bin/clang-cl.exe" -G Ninja -DCMAKE_BUILD_TYPE=Release
> cmake --build
> ```
> 
> 有一些陷阱，特别是你不能通过，因为这将触发LLVM 安装中`-DCMAKE_CXX_COMPILER=clang-cl.exe`的使用（这会导致第 1 点中提到的错误）。`clang-cl.exe`另外，忘记传递将会出错，并且由于 CMake 错误，`-DCMAKE_BUILD_TYPE=Release`您需要在路径中使用。`/`
> 
> 当涉及到修改时`CMAKE_CXX_FLAGS`，通常要避免这种情况。原因是该标志应该传递到命令行（手动或通过预设）。当您更改标志时，您会扰乱包管理器和下游消费者，因为您引入了不必要的或冲突的标志（或者在您的情况下，您完全覆盖了传递到命令行的任何内容）。这意味着下游消费者必须修补`CMakeLists.txt`，带来额外的维护负担。
> 
> _如果需要_一个标志才能编译工作（例如，如果没有此选项，项目将无法编译），您可以尝试使用语句检测确切的编译器，然后使用或`if()`将其添加到目标。“常见”示例包括和 链接器标志，例如.`target_compile_options(target PRIVATE /flag)``target_link_options``/bigobj``/STACK`



> I just tried building Triton `master` on Windows with ClangCL (latest Visual Studio update):
> 
> ```
> cmake -B build -T ClangCL "-DCMAKE_PREFIX_PATH:FILEPATH=d:\CodeBlocks\z3\install;d:\CodeBlocks\capstone\build\install;d:\CodeBlocks\llvm-12.0.0-rc1-install" -DLLVM_INTERFACE=ON
> cmake --build build --config Release
> ```
> 
> It works without any modifications to the `CMakeLists.txt` (or passing the CXX_STANDARD_REQUIRED variable). I saw this issue before, it happens in one of the following circumstances:
> 
> 1. Using a non-Microsoft version of `clang-cl` (eg downloaded from the latest LLVM releases, or modifying the VS installation)
> 2. Using CLion, where they made a major mistake and pass `-m64` to the clang command line, which completely breaks all builds. The issue was closed, but you can see it here: https://youtrack.jetbrains.com/issue/CPP-32063 (I am pretty confident it's not resolved yet, but perhaps it's only fixed in a prerelease).
> 
> To confirm everything works I also did a build with Ninja:
> 
> ```
> vs2022 x64 # open a VS2022 command line
> cmake -B build "-DCMAKE_PREFIX_PATH:FILEPATH=d:/CodeBlocks/z3/install;d:/CodeBlocks/capstone/build/install;d:/CodeBlocks/llvm-12.0.0-rc1-install" -DLLVM_INTERFACE=ON "-DCMAKE_C_COMPILER:FILEPATH=c:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/Llvm/x64/bin/clang-cl.exe" "-DCMAKE_CXX_COMPILER:FILEPATH=c:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/Llvm/x64/bin/clang-cl.exe" -G Ninja -DCMAKE_BUILD_TYPE=Release
> cmake --build
> ```
> 
> There are some pitfalls, notably you cannot pass `-DCMAKE_CXX_COMPILER=clang-cl.exe`, because this will trigger the usage of the `clang-cl.exe` in your LLVM installation (which causes errors are mentioned in point 1). Additionally forgetting to pass `-DCMAKE_BUILD_TYPE=Release` will error and you need to use `/` in the paths because of a CMake bug.
> 
> When it comes to modifying `CMAKE_CXX_FLAGS`, this is generally to be avoided. The reason is that this flag is supposed to be passed to the command line (either manually or by preset). When you are changing the flag you are messing with package managers and downstream consumers because you introduce unnecessary or conflicting flags (or in your case you completely override whatever is being passed to the command line). This means that downstream consumers have to patch the `CMakeLists.txt`, introducing additional maintenance burden.
> 
> If a flag is _required_ for the compilation to work (eg the project doesn't compile without this option), you try to detect the exact compilers using `if()` statements and then add it to the target with `target_compile_options(target PRIVATE /flag)` or `target_link_options`. "Common" examples of this include `/bigobj` and linker flags like `/STACK`.

ok, agree with your opinion.

cctv130:
> I just tried building Triton `master` on Windows with ClangCL (latest Visual Studio update):
> 
> ```
> cmake -B build -T ClangCL "-DCMAKE_PREFIX_PATH:FILEPATH=d:\CodeBlocks\z3\install;d:\CodeBlocks\capstone\build\install;d:\CodeBlocks\llvm-12.0.0-rc1-install" -DLLVM_INTERFACE=ON
> cmake --build build --config Release
> ```
> 
> It works without any modifications to the `CMakeLists.txt` (or passing the CXX_STANDARD_REQUIRED variable). I saw this issue before, it happens in one of the following circumstances:
> 
> 1. Using a non-Microsoft version of `clang-cl` (eg downloaded from the latest LLVM releases, or modifying the VS installation)
> 2. Using CLion, where they made a major mistake and pass `-m64` to the clang command line, which completely breaks all builds. The issue was closed, but you can see it here: https://youtrack.jetbrains.com/issue/CPP-32063 (I am pretty confident it's not resolved yet, but perhaps it's only fixed in a prerelease).
> 
> To confirm everything works I also did a build with Ninja:
> 
> ```
> vs2022 x64 # open a VS2022 command line
> cmake -B build "-DCMAKE_PREFIX_PATH:FILEPATH=d:/CodeBlocks/z3/install;d:/CodeBlocks/capstone/build/install;d:/CodeBlocks/llvm-12.0.0-rc1-install" -DLLVM_INTERFACE=ON "-DCMAKE_C_COMPILER:FILEPATH=c:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/Llvm/x64/bin/clang-cl.exe" "-DCMAKE_CXX_COMPILER:FILEPATH=c:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/Llvm/x64/bin/clang-cl.exe" -G Ninja -DCMAKE_BUILD_TYPE=Release
> cmake --build
> ```
> 
> There are some pitfalls, notably you cannot pass `-DCMAKE_CXX_COMPILER=clang-cl.exe`, because this will trigger the usage of the `clang-cl.exe` in your LLVM installation (which causes errors are mentioned in point 1). Additionally forgetting to pass `-DCMAKE_BUILD_TYPE=Release` will error and you need to use `/` in the paths because of a CMake bug.
> 
> When it comes to modifying `CMAKE_CXX_FLAGS`, this is generally to be avoided. The reason is that this flag is supposed to be passed to the command line (either manually or by preset). When you are changing the flag you are messing with package managers and downstream consumers because you introduce unnecessary or conflicting flags (or in your case you completely override whatever is being passed to the command line). This means that downstream consumers have to patch the `CMakeLists.txt`, introducing additional maintenance burden.
> 
> If a flag is _required_ for the compilation to work (eg the project doesn't compile without this option), you try to detect the exact compilers using `if()` statements and then add it to the target with `target_compile_options(target PRIVATE /flag)` or `target_link_options`. "Common" examples of this include `/bigobj` and linker flags like `/STACK`.

Yes, agree with your opinion

cctv130:
> I agreed with that we should avoid as much as possible touching the CMake file and let the user provide his arguments according to his setup.

I have a few questions here that came to me this afternoon:
LLVM lifting  will be able to vmprotect deobfuscation confused with themida etc have very good effect? Driver and application software in Windows platform？The latest information on 2023 is scarce, so please give us your opinion. Thank you.
Or is there a better way to do it.

cctv130:
@mrexodia One thing I need to explain is that I have no idea what compilation options are provided by llvm provided by Microsoft. And whether Microsoft provides the latest version in a timely manner is indeed a challenge for people who are new to llvm or who want to customize llvm.

cctv130:
@mrexodia And from what I've seen, the llvm provided by Microsoft doesn't seem to be able to compile LLVM pass.

cnheitman:
> I agreed with that we should avoid as much as possible touching the CMake file and let the user provide his arguments according to his setup.

I agree with this too. There seems to be no reason to modify the CMake file.

cctv130:
> I agreed with that we should avoid as much as possible touching the CMake file and let the user provide his arguments according to his setup.

The redundant cmake configuration has been removed, please check what else is wrong？

cctv130:
@JonathanSalwan 
And I found in your project, for this deobfuscation_process.png, help me get through the way of thinking, there is a big help to me, thank you very much.
image

cnheitman:
> The redundant cmake configuration has been removed, please check what else is wrong？

@cctv130 Why do you need to modify the files under `scripts/docker`?

cctv130:
@cnheitman The absolute path used for the environment variable related to py in this configuration is incorrect, and Meson cannot find it. It was stuck here for a long time during ci construction

cnheitman:
@cctv130 Ok, thanks. I will check that. It shouldn't be necessary to use conda in that script.

cctv130:
@cnheitman This mini conda and conda are different. Mini has a very small size. If you don't think it's necessary to add this configuration, I think this PR can closed. As for the script issue and the previous update of Bitwuzla to the latest version, you should submit the perfect PR. After all, I have a short time with Triton and need a lot of time to learn this framework. Sorry.

cnheitman:
@cctv130 Commit 357f372d953ab1aedca193f93ddd2165379ffd08 should fix the issue that you were having with the docker image. The problem was that the base image was updated (new python versions) and it broke the script. I'll rework the script so this doesn't happen again. For now the commit I mentioned should suffice.

cctv130:
@cnheitman thank you very much ^.^

PR #1286, opened 2023-09-13
Upgrade Bitwuzla to version 0.2.0

cnheitman:
This PR upgrades Bitwuzla to version 0.2.0.

The only remaining file to update is `Dockerfile` which will upgrade once `master` syncs with `dev-v1.0`.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -48,7 +48,7 @@ jobs:
         run: |
           git clone https://github.com/bitwuzla/bitwuzla.git
           cd bitwuzla
-          git checkout -b 0.1.0 0.1.0
+          git checkout -b 0.2.0 0.2.0
           python ./configure.py --shared
           cd build
           sudo ninja install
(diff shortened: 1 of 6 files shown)


JonathanSalwan:
do you think we can do some `#if #else` to support v0.1 and v0.2? Like [we do](https://github.com/JonathanSalwan/Triton/blob/master/src/libtriton/arch/arm/aarch64/aarch64Specifications.cpp#L357) for capstone? So that we can support both versions and not force users to upgrade.

cnheitman:
Bitwuzla does not seem to provide a `BZLA_VERSION` define in their includes. We can propose it on their repo in for handling that in the future. Regarding this particular PR, it does not seem to have change much (at least, from our standpoint). I'm not sure they are going to keep releasing as often as they did with v0.1.0 and now v0.2.0. They did a big refactoring of the entire project and it seems to be stable (or soon to be).

JonathanSalwan:
Ack, no problem :).

PR #1288, opened 2023-09-23
Dev v1.0

cctv130:
add linux auto copy libz3.so  to the build_lib directory.


cctv130:
@JonathanSalwan Fixed, please check.

PR #1289, opened 2023-09-29
Fix Bitwuzla VARIABLE_NODE translation

cnheitman:
This PR fix an issue with the in the Triton to Bitwuzla translation for `VARIABLE_NODE`. If a variable node appears more that once in an expression, different term for that node will be created. That means, `bitwuzla_mk_const` will return different `BitwuzlaTerm`s for the same symbol. The fix checks whether the term was already created and returns it, otherwise creates a new one.


JonathanSalwan:
Is it an old bug or is it due to the new bitwuzla version?

cnheitman:
I think this also happens in the old version of Bitwuzla but I haven't checked.

Issue #1136, opened 2022-06-10
Python bindings for `tritonToZ3` segfault


jordan9001:
```py
from triton import *

ctx = TritonContext(ARCH.X86_64)

astctx = ctx.getAstContext()

ast = astctx.bvtrue()

astctx.tritonToZ3(ast)
```

The above can produce a crash when the python bindings do the `Z3_translate`

I think it might have to do with how the `z3.main_ctx().ctx.value` context is used?

JonathanSalwan:
Hey,

Is your `libz3.so` is the same that used for the compile and link of the `libtriton.so` ?

Also maybe related to: https://github.com/JonathanSalwan/Triton/issues/1024

jordan9001:
Hmm, I think it is related to that issue. Thank you

```py
import triton

# in maps:
# /usr/lib64/libz3.so.4.8.14.0

import z3

# in maps:
# /mnt/A/jay_home/.local/lib/python3.9/site-packages/z3/lib/libz3.so
# /usr/lib64/libz3.so.4.8.14.0
```

JonathanSalwan:
Is `~/.local/lib/python3.9/site-packages/z3/lib/libz3.so` the same that `/usr/lib64/libz3.so.4.8.14.0` ? Also, make sure that Triton used at the compile the include of z3 4.8.14.0 (if you have multiple z3 versions).

jordan9001:
No, the file is different, not symlink'd. But both are version `4.8.14.0` though.

cnheitman:
I had this issue before. You have to make sure that `triton.so` uses `~/.local/lib/python3.9/site-packages/z3/lib/libz3.so` (either by pointing to it when compiling Triton with `-DZ3_LIBRARIES=~/.local/lib/python3.9/site-packages/z3/lib/libz3.so` or setting `LD_LIBRARY_PATH=~/.local/lib/python3.9/site-packages/z3/lib` so it uses the correct one when it loads).

PR #1296, opened 2023-12-04
Fix Z3 dependency inconsistency and missing dependencies in Dockerfile

m-carrasco:
Hi :wave: 

I've noticed some difficulties while building Triton from sources using the available Dockerfile. This PR aims to fix them.

1. Now, Z3 is installed only once using pip to get rid of crashes like #1136. This also decreases the time taken to build the Docker image.
2. The test suite was not executed for the Docker image, which is now.
3. Some dependencies were missing for the test suite to run successfully.
4. The Dockerfile was cloning the source code. This prevented us from testing the changes in the CI when working on a branch other than master.

Best regards,
Manuel.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -7,5 +7,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+    - name: Set Timestamp Variable
+      run: echo "TIMESTAMP=$(date +%s)" >> $GITHUB_ENV
     - name: Build the Docker image
-      run: docker build . --file Dockerfile --tag triton-docker:$(date +%s)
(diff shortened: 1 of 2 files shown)


m-carrasco:
The Docker image is built successfully and all test cases are good.

I'm not sure if the remaining failures are triggered by my changes. I missed setting version 4.8.14 for Z3 as in master, instead of the latest release 4.12.2. Could this be triggering the errors?

JonathanSalwan:
I think those errors are not coming from your commit.

m-carrasco:
Thank you @JonathanSalwan 

I've just pushed a commit restoring the previous Z3 version. Let me know if I could do anything else.

cnheitman:
Thanks @m-carrasco . It looks like the failures are not related to your commit. I'll merge your request (and take a look at CI to fix those other issues).

PR #1297, opened 2023-12-06
Update CI

cnheitman:
This PR:

- Solves issues with some workflows (codecov, linux, and vcpkg). The issues related to a missing package requiered by `z3-solver`.
- Upgrades the dockerfile and script that are used to build the Python package for Linux. All dependencies have been upgraded to their latest (or close to) versions. Also, it moves from `manylinux_2_31_x86_64` to `manylinux_2_28_x86_64`.
- Upgrades the C++ version that is used to build Triton from 14 to 17. This allows to compile Triton against LLVM 16.
- Updates the python workflow. The `build-windows` job has been reintroduce (although, it does not include LLVM as there is no pre-compiled package for Windows). The `build-macos` job has been updated to use the latest versions of the dependencies (where possible).


JonathanSalwan:
Awesome, thx Chris! I'm merging this one. Regarding Appveyor, it looks like they changed something in their build system which makes the LIEF wheel invalid.

JonathanSalwan:
Regarding CI on Appveyor, I've removed the matrix for Python3.6 and 3.7 as they are not maintained anymore (https://endoflife.date/python). However, we should now support 3.11 and 3.12 in our CI.

PR #1315, opened 2024-03-24
Added disassembly by filterCallback

Japrajah:


Japrajah:
Because it is called before buildSemantics, it is only able to filter by string & controlFlow, and isMemoryRead does not work.
But if you can call buildSemantics, you can use everything. But it will be equivalent to call 'processing'.
```
   static triton::Context* hack = &ctx;
    const  auto  dism_filter = [](std::vector<triton::arch::Instruction>& ret) -> bool
        {
            auto instr = ret.back();
            hack->buildSemantics(instr); // will not work without this
             return   instr.isMemoryRead() && !instr.isReadFrom(hack->registers.x86_rsp); 
        };
    ctx.disassembly(0x40000, dism_filter);
       const  auto  dism_filter_fine = [](std::vector<triton::arch::Instruction>& ret) -> bool
       {
           auto instr = ret.back();
           return  instr.getDisassembly().contains("rdrand"); 
       };
          ctx.disassembly(0x40000, dism_filter_fine );
```


Should there be a function that processes until "something", because now we need to dissasemble everything, that's can be unused latter.

PR #1322, opened 2024-05-06
Fix addCallback binding

cnheitman:
Fix PR fixes a small memory leak in the `addCallback` function. This issue was spotted than to LIEF and their new nanobind implementation of its Python bindings (ref [here](https://nanobind.readthedocs.io/en/latest/faq.html#why-am-i-getting-errors-about-leaked-functions-and-types)).

### DIFF
--- a/src/libtriton/bindings/python/objects/pyTritonContext.cpp
+++ b/src/libtriton/bindings/python/objects/pyTritonContext.cpp
@@ -503,13 +503,10 @@ namespace triton {
         if (PyMethod_Check(function)) {
           cb_self = PyMethod_GET_SELF(function);
           cb = PyMethod_GET_FUNCTION(function);
-
-          Py_INCREF(cb_self);
         }
         else {
           cb = function;
         }
-        Py_INCREF(cb);
 
         try {
           switch (static_cast<triton::callbacks::callback_e>(PyLong_AsUint32(mode))) {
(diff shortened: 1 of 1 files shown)


JonathanSalwan:
Mmmh, I remember adding these `INC_REF` due to a bad counting. I think that's why unit tests segfault =/

cnheitman:
I'm checking the tests to understand better what's going on.

JonathanSalwan:
Here is the commit that added these `INC`. Maybe it can help you. https://github.com/JonathanSalwan/Triton/commit/23f5c21db414513194ace26358b386fc409e24bb

cnheitman:
I updated the PR. Now tests pass (at least locally), and I'm not getting the nanobind leaked messages when using Triton with LIEF v0.14.1. It does not seem necessary to increase the refs in `addCallback` as the are "borrowed" refs and, in a similar way, they don't need to be decreased in `removeCallback`.

Let's wait until all tests run and see what happens.

jordan9001:
These changes brought back this SEGFAULT

https://github.com/JonathanSalwan/Triton/issues/1035

It seems like we need a better fix than before. We need some way to know when we will require a reference due to a nested function.

JonathanSalwan:
Erf, you're right and I don't know why we didn't have this unittest... @cnheitman can you backport the patch and try another fix for the memleak?

jordan9001:
I think it makes sense to me that we require the addition of a reference to any function object sent to `addCallback` (and subtraction on `removeCallback`).

Could the cause of the memory leak be that `removeCallback` isn't always being called when the context is destroyed? Do we need to add logic to `TritonContext_dealloc` to remove any remaining references?

JonathanSalwan:
I've revert #1322 and added an unit test for #1035. Now, let's see how we can fix the memory leak.

cnheitman:
Hey! A bit late... I'll try to check this again (and how it interacts with #1035) later this week.

PR #1323, opened 2024-05-09
Fix addCallback binding

cnheitman:
This PR is related to PR #1322. It makes sure to release the args object in case the callback throws an exception.


JonathanSalwan:
Nice catch !

PR #1330, opened 2024-05-26
Add the XMM16-31 and YMM16-31 registers for x86_64

jordan9001:
This adds the missing registers used to index into the lower portions of the AVX-512 registers. Here I have only added the registers, as well as adjust the get/set concrete register value function to use the correct parent registers. It is a simple change, and adds support for instructions using these registers.

### DIFF
--- a/src/libtriton/arch/x86/x8664Cpu.cpp
+++ b/src/libtriton/arch/x86/x8664Cpu.cpp
@@ -331,7 +331,7 @@ namespace triton {
 
 
       bool x8664Cpu::isSSE(triton::arch::register_e regId) const {
-        return ((regId >= triton::arch::ID_REG_X86_MXCSR && regId <= triton::arch::ID_REG_X86_XMM31) ? true : false);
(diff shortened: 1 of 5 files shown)


src/libtriton/arch/x86/x8664Cpu.cpp
```diff
@@ -331,7 +331,7 @@ namespace triton {
 
 
       bool x8664Cpu::isSSE(triton::arch::register_e regId) const {
-        return ((regId >= triton::arch::ID_REG_X86_MXCSR && regId <= triton::arch::ID_REG_X86_XMM15) ? true : false);
```
JonathanSalwan, line 334: I'm not sure about this. I think SSE only extends registers from xmm0 to xmm15? xmm16-31 are introduced from AVX (EVEX)?

* https://en.wikipedia.org/wiki/AVX-512
* https://en.wikipedia.org/wiki/EVEX_prefix

JonathanSalwan:
Thanks for this MR. Only one remark about `isSSE`. Other than that, everything looks good to me :)

PR #1331, opened 2024-05-28
implement Aarch64 UMOV instruction semantics

psi-func:
add more VAS types to correctly handle different cases which part of vector register should move to GPR register
add umov semantics in aarch64
instruction reference [here](https://developer.arm.com/documentation/dui0801/l/A64-SIMD-Vector-Instructions/UMOV--vector---A64-)

### DIFF
--- a/src/libtriton/arch/arm/aarch64/aarch64Semantics.cpp
+++ b/src/libtriton/arch/arm/aarch64/aarch64Semantics.cpp
@@ -5,7 +5,6 @@
 **  This program is under the terms of the Apache License 2.0.
 */
 
-#include <utility>
 #include <triton/aarch64Semantics.hpp>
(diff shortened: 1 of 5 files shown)


psi-func:
Capstone v4.0.1 missing important feature with parsing extended VAS types
build fails in CI

gulldan:
also need it ASAP

JonathanSalwan:
Lol, I think this is the first ever MR that involves so many reactions ^^.

* Regarding Capstone 5: As 5.0.1 is now tagged as release, I think we can drop support for 4.x.
* Regarding unit tests. Indeed, it could be great to add some in https://github.com/JonathanSalwan/Triton/blob/master/src/testers/aarch64/unicorn_test_aarch64.py

PR #1332, opened 2024-05-29
Add tests for umov aarch64 instructions

psi-func:
+ Remove capstone 4.0.1 from CI


psi-func:
sorry, `#ifdef` with capstone major version makes some instructions run incorrectly
+ add some tests for cmp unicorn and triton

psi-func:
Wrote comment in previous [PR](https://github.com/JonathanSalwan/Triton/pull/1331#issuecomment-2137178057)
Found bug in capstone
Can fix it with additional logic in mov_s function semantics

JonathanSalwan:
I think you should rebase on `dev-v1.0` and just push the fix of your `umov` semantic. Otherwise, we will have conflicts merging this one.

psi-func:
> I think you should rebase on `dev-v1.0` and just push the fix of your `umov` semantic. Otherwise, we will have conflicts merging this one.

PR #1333, opened 2024-05-30
More Aarch64 instruction support

psi-func:
added tests for new instructions
drop capstone 4.0.1 support


psi-func:
Found more bugs in capstone for aarch64
It can't disasm fmov normally when it use vector indices
checked with capstone next branch -> ok

Issue #1335, opened 2024-06-13
getParentRegisters() returns some sub-registers


jordan9001:
The `getParentRegisters()` function for x86_64 returns some duplicate registers, specifically `isSSM`, `isAVX256` `isAVX512` are all included.
I am happy to submit a pull request, but I was would like your feedback first. It makes sense to me that the only parent registers listed on x86 should be the YMM registers, none of the `isSSM` registers. On x86_64 it should just be the zmm registers. Do you agree?

Essentially, I would like to add this test to the tests:

```py
class TestRegisterParents(unittest.TestCase):
    """Test register Parent Register List"""

    def setUp(self):
        """Define the arch list"""
        self.archctx = []
        self.archcxt.append(TritonContext(ARCH.X86))
        self.archcxt.append(TritonContext(ARCH.X86_64))
        self.archcxt.append(TritonContext(ARCH.ARM32))
        self.archcxt.append(TritonContext(ARCH.AARCH64))

    def test_reg_parents(self):
        for ctx in self.archctx:
            for pr in ctx.getParentRegisters():
                self.assertEqual(pr, ctx.getParentRegister(pr))
```

I would probably also add and use a `isAVX512Parent` function, because the `isAVX512` function returns some xmm and ymm identifiers as well.

jordan9001:
It looks like AARCH64's `getParentRegisters()` lists the `d#` registers but not the `q#` registers. So, if it makes sense I could replace those as well.

JonathanSalwan:
Thx for this report. Indeed, it would make more sense. Sounds good to me :)

jordan9001:
@JonathanSalwan Is there a specific reason some flag registers are their own parent? Or is it okay if I have ZF's parent be EFLAGS and EFER_TCE's parent be EFER?

PR #1336, opened 2024-06-15
Ensure getParentRegisters returns all the parent register.

jordan9001:
Adds a new test and updates the getParentRegisters implementations so that all parent register are returned, and sub-registers are not.
See https://github.com/JonathanSalwan/Triton/issues/1335

### DIFF
--- a/src/libtriton/arch/arm/aarch64/aarch64Cpu.cpp
+++ b/src/libtriton/arch/arm/aarch64/aarch64Cpu.cpp
@@ -297,20 +297,28 @@ namespace triton {
             auto regId = kv.first;
             const auto& reg = kv.second;
 
-            /* Skip Vector and System registers */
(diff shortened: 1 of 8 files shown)


src/libtriton/arch/arm/aarch64/aarch64Cpu.cpp
```diff
@@ -297,21 +297,29 @@ namespace triton {
             auto regId = kv.first;
             const auto& reg = kv.second;
 
-            /* Skip Vector and System registers */
-            if (this->isVectorRegister(regId) || this->isSystemRegister(regId))
```
jordan9001, line 301: Why were the system registers being skipped originally? Is there a reason they should not be included?

JonathanSalwan: If i remember correctly it was because I did not want to pollute to much the output. So, clearly not a good reason ^^.

PR #1349, opened 2024-08-08
Switch to Capstone 5.x

### LINKED ISSUE #1348: Incorrect CMake dependency management
Triton uses its own CMake scripts to search for dependencies, moreover, variables with the CS_VERSION_MAJOR project version are created for capstone, on which the project build further depends
If Triton is used together with the package manager (conan), the search can be changed automatically and custom scripts will not be run, there will be no variables and, for example, this condition will break CMake

```cmake
if(${CS_VERSION_MAJOR} GREATER_EQUAL 5)
```
Screenshot from 2024-08-08 16-30-42

mrexodia:
Definitely requires some work to fix the CI, but I will work on that in my free time. After this is merged the AppVeyor file can also be deleted.

Closes #1348


mrexodia:
So funnily enough now only the Windows VCPKG build succeeds? I upgraded LIEF to the latest version, but for some reason that doesn't seem to work? 🤔

mrexodia:
Looks like CI passed for all platforms (including Windows), so you can merge whenever @JonathanSalwan. I removed the appveyor file, but it still shows a failing check for some reason...

PR #1359, opened 2024-09-02
Aarch64 floating point operands support

psi-func:
Implemented initial support for Aarch64 float-point operands in instructions
tested on `FMOV`, other instructions with this feature are not implemented in triton (float point nature)


JonathanSalwan:
Thx!

What about moving `encode_fp_imm` into the `Immediate` class as a new constructor so that the other architectures can benefit of this feature?

Something like this:

```
Immediate::Immediate(double value, triton::uint32 size) {
    this->setValue(this->encode_fp_imm(value), size);
}
```

Then, at this [line](https://github.com/JonathanSalwan/Triton/pull/1359/files#diff-f7d5a33e95a33760b0164cf54cb29eec287f35518d205706c97ef2ebf39c0506R588) you can just call the new constructor as follow: `Immediate imm{op->fp, size};`.

psi-func:
I agree, move feature to Immediate ctor + add endianness check

PR #1368, opened 2024-10-13
Fix unsupported instruction for some bics variant

llaubin:
hola o/ 

Just a tiny fix to address some variant of bics instruction, like `bics wzr, w23, w17`, which were not recognized. 

Lo;

### DIFF
--- a/src/libtriton/arch/arm/aarch64/aarch64Specifications.cpp
+++ b/src/libtriton/arch/arm/aarch64/aarch64Specifications.cpp
@@ -405,6 +405,7 @@ namespace triton {
               break;
 
             case triton::extlibs::capstone::ARM64_INS_BIC:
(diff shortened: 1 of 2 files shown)


src/libtriton/arch/arm/aarch64/aarch64Specifications.cpp
```diff
@@ -405,6 +405,7 @@ namespace triton {
               break;
 
             case triton::extlibs::capstone::ARM64_INS_BIC:
+            case triton::extlibs::capstone::ARM64_INS_BICS:
```
cnheitman, line 408: In principle, this should be added inside a `#if CS_API_MAJOR >= 5` guard, as in the case of `ADDS` and `ANDS`, just to keep the same style. On the other hand, we have already move to Capstone >= 5.0.x, so I don't know whether it is worth to do anymore.

PR #1382, opened 2024-12-23
Update Python versions

cnheitman:
This PR updates the Python versions of some of the CI, and drops support for Python 3.8 as the end of life was 2024-10-07.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.9]
+        python-version: [3.11]
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


mrexodia:
Isn't it a bit early to drop support for 3.8, it's been 2 months 😅

cnheitman:
Hey! Too early? :sweat_smile: 

Keep in mind this change only affects the CI. No change was made to the code, so it still builds for Python 3.8 locally.

Mostly, it is for simplifying the maintenance, as we already support 5 versions of Python.

I could postpone this change if necessary. Are you aware of anyone affected too much by this change?

mrexodia:
I opened #1383, which would solve this issue once and for all (more as an fyi). As for the impact, it would remove the binary wheels for python 3.8 from pypi right?

cnheitman:
> I opened https://github.com/JonathanSalwan/Triton/issues/1383, which would solve this issue once and for all (more as an fyi).

Thanks, I've just replied to your issue.

> As for the impact, it would remove the binary wheels for python 3.8 from pypi right?

Yes, in principle, I will stop pushing the py38 wheel to PyPI.

PR #1386, opened 2025-01-11
fix + improve python type stubs

Toizi:
There were a lot of errors in the generated type stub file due to invalid assignments from the namespace generation and type errors using the `self.org` forward approach. The stub file now uses enums to deal with the namespaces and only sets the docstring and type information in the functions.

The new code now also tries to handle overloads by merging functions since overloads with the same name are not allowed, e.g.
```python
    def disassembly(self, inst):
        # type: (Instruction) -> None
        """Disassembles the instruction and sets up operands."""
        return self.org.disassembly(inst)


    def disassembly(self, block, addr=0):
        # type: (BasicBlock, int) -> None
        """Disassembles a basic block with a potential given base address."""
        return self.org.disassembly(block, addr=0)


    def disassembly(self, addr, count):
        # type: (int, int) -> List[Instruction]
        """Disassembles a concrete memory area from `addr` and returns a list of at most `count` disassembled instructions."""
        return self.org.disassembly(addr, count)


    def disassembly(self, addr):
        # type: (int) -> BasicBlock
        """Disassembles a concrete memory area from `addr` to control flow instruction and returns a BasicBlock."""
        return self.org.disassembly(addr)

```

turned into
```python
    def disassembly(self, block_inst_addr, addr_count):
        # type: (Self, int | BasicBlock | Instruction, int) -> None
        """Disassembles the instruction and sets up operands.

Disassembles a basic block with a potential given base address.

Disassembles a concrete memory area from `addr` and returns a list of at most `count` disassembled instructions.

Disassembles a concrete memory area from `addr` to control flow instruction and returns a BasicBlock."""
        pass

```
This is not always a perfect solution depending on the overloads. From what I have seen it is basically always better than having just one because the others are ignored though.

Attached is the generated stub file in case you want to compare it to the current one. Just rename it to `.pyi` since github does not allow arbitrary file extensions.
[triton.txt](https://github.com/user-attachments/files/18384620/triton.txt)

### DIFF
--- a/src/libtriton/bindings/python/namespaces/initStubsNamespace.cpp
+++ b/src/libtriton/bindings/python/namespaces/initStubsNamespace.cpp
@@ -28,28 +28,28 @@ The STUBS namespace contains all stubs.
 \section STUBS_py_api Python API - Items of the STUBS namespace
 <hr>
 
-- **STUBS.AARCH64.LIBC.code**<br>
+- <b>STUBS.AARCH64.LIBC.code</b><br>
 The libc stub on AArch64 architecture with the ARM64 ABI calling convention.
 
-- **STUBS.AARCH64.LIBC.symbols**<br>
+- <b>STUBS.AARCH64.LIBC.symbols</b><br>
 The symbols map of the AArch64 libc stub.
 
-- **STUBS.I386.SYSTEMV.LIBC.code**<br>
+- <b>STUBS.I386.SYSTEMV.LIBC.code</b><br>
 The libc stub on i386 architecture with the SystemV ABI calling convention.
 
-- **STUBS.I386.SYSTEMV.LIBC.symbols**<br>
+- <b>STUBS.I386.SYSTEMV.LIBC.symbols</b><br>
 The symbols map of the i386-systemv libc stub.
 
-- **STUBS.X8664.MS.LIBC.code**<br>
+- <b>STUBS.X8664.MS.LIBC.code</b><br>
 The libc stub on x8664 architecture with the MS ABI calling convention.
 
-- **STUBS.X8664.MS.LIBC.symbols**<br>
+- <b>STUBS.X8664.MS.LIBC.symbols</b><br>
 The symbols map of the x8664-ms libc stub.
 
-- **STUBS.X8664.SYSTEMV.LIBC.code**<br>
+- <b>STUBS.X8664.SYSTEMV.LIBC.code</b><br>
 The libc stub on x8664 architecture with the SystemV ABI calling convention.
 
-- **STUBS.X8664.SYSTEMV.LIBC.symbols**<br>
+- <b>STUBS.X8664.SYSTEMV.LIBC.symbols</b><br>
 The symbols map of the x8664-systemv libc stub.
 
 */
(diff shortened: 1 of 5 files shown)


src/libtriton/bindings/python/namespaces/initStubsNamespace.cpp
```diff
@@ -28,28 +28,28 @@ The STUBS namespace contains all stubs.
...
 
-- <b>STUBS.X8664.MS.LIBC.symbols</b><br>
+- **STUBS.X8664.MS.LIBC.symbols**<br>
 The symbols map of the x8664-ms libc stub.
 
-- <b>STUBS.X8664.SYSTEMV.LIBC.code</b><br>
+- **STUBS.X8664.SYSTEMV.LIBC.code**<br>
 The libc stub on x8664 architecture with the SystemV ABI calling convention.
 
-- <b>STUBS.X8664.SYSTEMV.LIBC.symbols</b><br>
```
JonathanSalwan, line 52: Can you keep this file like it was? It's for Doxygen web page generation.

JonathanSalwan (commented):
Only one comment. Other than that, lgtm. Thx!

src/libtriton/bindings/python/namespaces/initStubsNamespace.cpp
bearpebble: @JonathanSalwan I was under the impression that `**` would be markdown for bold, i.e. the same as the `<b>` tag. It's used in all other namespaces and the type generation code matches these to find the names.
If I change them here I'll either need to adjust the code or they won't be part of the type stubs.

Are you sure this needs the `<b>` tags?

Toizi: @JonathanSalwan I was under the impression that `**` would be markdown for bold, i.e. the same as the `<b>` tag. It's used in all other namespaces and the type generation code matches these to find the names.
If I change them here I'll either need to adjust the code or they won't be part of the type stubs.

Are you sure this needs the `<b>` tags?

JonathanSalwan: So, why are you keeping this `<b>` here: https://github.com/JonathanSalwan/Triton/pull/1386/files#diff-3916adb03cd698380dda94ae3005c0ef0de07fc7ff37a6acfa49f22399d37b8aR301 ? I think we are using `<b>` all over the doc ?

Toizi: I just checked and there no uses of the `<b>` tag in the namespaces directory so far. `<b>` was already matched on the other directories though.

I ended up adding `<b>` for namespaces since this might happen in the future again and reverted the doc changes like you requested.

JonathanSalwan: Erf, indeed you're right. Looks like the tag `<b>` is only present for modules (i do not remember why it's like this). Anyway, it's a good thing to support both tags: `**` and `<b>`. Do you think we should keep `**` for namespace in order to get a feeling of consistency ?

Toizi: I'd keep it as is. Sooner or later somebody will end up using `<b>` anyway if it's not enforced.

PR #1392, opened 2025-02-14
Update Unicorn

cnheitman:
This PR updates the CI code to use the latest version of Unicorn.

Unicorn v2.1.1 does not support some of the tested instructions: `pextr`, `vpextr`, `pinsr`. There are also some issues with the following instructions: `lzcnt`, and `tzcnt`. Therefore, tests for those instructions were commented. It is possible to use Unicorn v2.0.0 in order to test them locally, if needed. Currently we moved to v2.1.1 since previous versions do not work on newer OSX versions (>= v14), and that's why the OSX workflow is currently failing.

@JonathanSalwan Let me know if you agree with this change.


JonathanSalwan:
Yeah I was also looking at this issue today... and thought about something like testing the `unicorn.__version__`. e.g:

```
CODE  = [
     ("2.0.1", None, [], b"\x66\x0f\x3a\x14\xc8\x00", "pextrb eax, xmm1, 0"),
    ]
```

However, i don't know if it's worth keeping old versions...

JonathanSalwan:
It's surprising how much instructions are no more supported by the new UC version tho...

cnheitman:
Yeah, I don't know what's going on with Unicorn, and why some instrs do not longer work...

Regarding maintaining compatibility with older versions, it seems that it brings more maintenance effort than the benefit of it. 

For now we have issues with only those 5 instrs for x86, and we are confident enough that they are implemented correctly. So, my opinion is to move on to the latest version of Unicorn and if we see that other issues start to appear we can think of supporting multiple versions.

JonathanSalwan:
Yeah ok i'm good with that 👍

PR #1410, opened 2025-07-01
Add 128-bit integer support in tritonToLLVM

raax7:
Extended tritonToLLVM.cpp to support 128-bit integers.
- Expanded BV_NODE to handle 128-bit bit-vectors
- Added switch cases to argument type generation and BSWAP_NODE to select 128-bit integer types


raax7:
During testing, I discovered a bug caused by certain ASTs that Triton can emit, such as: 
`((_ extract 127 0) (concat ((_ extract 511 128) (_ bv0 512)) xmm0))`
This caused an exception to be thrown since the code only handled BV_NODEs with bit-widths less than 64-bits, or exactly 128-bits.
The fix for this issue I implemented is simply to extend BV_NODEs to handle bit-widths up to 512-bits.

JonathanSalwan:
nop, looks good to me as well. thx @cnheitman and @raax7

JonathanSalwan:
btw, in our CI, i think we should also test for llvm versions newer than just llvm-14

PR #1413, opened 2025-07-21
fix missing opcode bindings

redthing1:
a number of opcode bindings were missing for python. i added them.


JonathanSalwan:
I've changed the base dst branch, there is only one conflict to fix with `initOpcodesNamespace.cpp`

redthing1:
conflict is fixed

PR #1416, opened 2025-08-05
add cstdint include

KarlVogel:
Fix for compilation issue (ie. on fedora 42 with gcc-15):

``` 
src/libtriton/includes/triton/modesEnums.hpp:51:64: error: ‘uint64_t’ was not declared in this scope
```


JonathanSalwan:
shouldn't be better to add the include in `modesEns.hpp` instead of `modes.c`?

PR #1417, opened 2025-08-05
Don't trigger a callback when symbolizing memory

KarlVogel:
Using `symbolizeMemory(mem, "label")` calls `setConcreteMemoryValue()` which then triggers the `SET_CONCRETE_MEMORY_VALUE` callback. When tracking writes from an app, this would then show up as a write to memory, but since it's the framework doing it, it doesn't make sense to trigger the callback.

NOTE: This issue might be debatable if the callback should actually trigger or not. In my use case, it was interfering as it showed up as a "ghost" write, which the application being monitored didn't actually perform.


JonathanSalwan:
mmmh, actually I'm reverting this patch as we have some cases that need to trigger the callback. I also understand your point. So, the best solution would be to get an optional arg to decide whether we call or not the callback.

`symbolizeMemory(mem, "label", callback=true)`.

PR #1449, opened 2026-05-18
fix: cmpxchg when dest=al

### LINKED ISSUE #1448: x64: CMPXCHG when DEST=AL
The [CMPXCHG](https://www.felixcloutier.com/x86/cmpxchg) instruction compares the destination operand with AL/AX/EAX/RAX and if they are equal, sets the destination operand to the source operand.

```text
(* Accumulator = AL, AX, EAX, or RAX depending on whether a byte, word, doubleword, or quadword comparison is being performed *)
TEMP := DEST
IF accumulator = TEMP
    THEN
        ZF := 1;
        DEST := SRC;
    ELSE
        ZF := 0;
        accumulator := TEMP;
        DEST := TEMP;
FI;
```

However, in Triton, if the destination operand is `al` then its value never changes

```py
from triton import *

context = TritonContext(ARCH.X86_64)

# cmpxchg al, bl
instruction = Instruction(0x1000, b"\x0f\xb0\xd8")

# set the register values
context.setConcreteRegisterValue(context.registers.rax, 0)
context.setConcreteRegisterValue(context.registers.rbx, 1)

context.processing(instruction)

rax = context.getConcreteRegisterValue(context.registers.rax)
rbx = context.getConcreteRegisterValue(context.registers.rbx)

print(f"rax: {rax:x}, rbx: {rbx:x}")  # rax: 0, rbx: 1
```

---

This was tested on the python bindings for triton: `triton-library=1.0.0rc4`

Garfield1002:
Fixes #1448

### DIFF
--- a/src/libtriton/arch/x86/x86Semantics.cpp
+++ b/src/libtriton/arch/x86/x86Semantics.cpp
@@ -5051,9 +5051,7 @@ namespace triton {
           expr6 = this->symbolicEngine->createSymbolicExpression(inst, node2, src1, "XCHG operation");
 
         /* Accumulator */
-        if (nodeq->evaluate() == true)
-          expr7 = this->symbolicEngine->createSymbolicExpression(inst, node3p, accumulatorp, "XCHG operation");
-        else
+        if (nodeq->evaluate() == false)
           expr7 = this->symbolicEngine->createSymbolicExpression(inst, node3, accumulator, "XCHG operation");
 
         /* Spread taint */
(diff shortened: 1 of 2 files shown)


JonathanSalwan:
Can you add some unit tests in https://github.com/JonathanSalwan/Triton/blob/master/src/testers/x86/unicorn_test_x86.py ? See an [example](https://github.com/JonathanSalwan/Triton/blob/master/src/testers/x86/unicorn_test_x86.py#L962-L965). You just have to add a small snippet of code somewhere in the `CODE` list. We will emulate this with triton and unicorn to check mismatch.

PR #1452, opened 2026-06-24
Migrate off deprecated LLVM pointer and interinsic APIs

321Proteus:
Because `llvm::Intrinsic::getDeclaration()` was deprecated in LLVM 20 and was removed in [this pull](https://github.com/llvm/llvm-project/pull/152645), I was unable to build the lifter with the latest LLVM in my system. Replaced the calls with the new `getOrInsertDeclaration()`.

Also updated the pointers in `SELECT_NODE` and `STORE_NODE` in order to get rid of the deprecated `llvm::Type::getPointerTo()` ([also from LLVM 20](https://github.com/llvm/llvm-project/pull/113331)) in favour of the opaque pointers (`llvm::PointerType::get()`).

### DIFF
--- a/src/libtriton/ast/llvm/tritonToLLVM.cpp
+++ b/src/libtriton/ast/llvm/tritonToLLVM.cpp
@@ -151,6 +151,7 @@ namespace triton {
 
         case triton::ast::BSWAP_NODE: {
           llvm::Function* bswap = nullptr;
+#if LLVM_VERSION_MAJOR >= 20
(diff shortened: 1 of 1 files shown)


321Proteus:
The check builds use LLVM v14 so will fail no matter what

mrexodia:
Just add some #ifdef based on the LLVM version to support both

321Proteus (commented):
LLVM preproc def has been successfully updated and should pass checks now

PR #1456, opened 2026-08-27
Run the Unicorn semantics tests on Windows too

mahmoudimus:
> **Stacked on #1457 and #1455.** The first two commits here are those PRs; please review
> only the third. #1457 fixes a pre-existing CMake 4 failure in the vcpkg workflow;
> #1455 is the bug this PR's test change uncovers.

### Summary

`src/testers/CMakeLists.txt` has gated the 39 semantics tests to Linux and Darwin since they were added, with the comment *"because the setup is easier."* That is no longer true, and the gate was concealing a real Windows-only correctness bug.

### Why the reason no longer holds

* `unicorn` (2.1.4) and `lief` (1.0.0) both publish `win_amd64` wheels.
* None of the 27 `unicorn_test*.py` scripts or the 12 lief-driven ARM32 crypto runners contain anything platform specific. Searching all of them for `mmap`, `fcntl`, `subprocess`, `os.uname` and `platform.system` matches only the `#!` line, and `ctest` invokes them through `${PYTHON_EXECUTABLE}` rather than the shebang.
* `.github/actions/install-triton-deps/windows` already runs `python -m pip install unicorn lief`, so Windows CI has been paying for the dependencies without using them.

### Measured

`windows-2022` as subject, `ubuntu-24.04` as control on an identical configuration — capstone 5.0.7, unicorn 2.1.4, lief 1.0.0, `LLVM_INTERFACE=OFF` on both, versions confirmed from the job logs:

| Leg | Result | Time |
|---|---|---|
| windows-x64 | 40 / 41 | 605.6s |
| ubuntu-24.04 (control) | 47 / 47 | 526.2s |

The totals differ because Windows does not register the C++ example tests and adds `DummyTest`.

38 of the 39 previously-gated tests pass on Windows unchanged. The single failure was:

```
[KO] fmov s0, #2.0
	v0: 0x40000000 (UC) != 0x40 (TT)
```

which is the bug fixed by #1455. With it applied, both platforms are green.

### Cost and benefit

Roughly ten minutes of Windows CI time. In exchange, the AArch64, ARM32, RISC-V and x86 semantics get cross-validated against a second emulator on Windows.

The Python unit tests cannot substitute for this. They check fixed expected values, so they only catch what someone thought to write down — `test_semantics.py` and `test_emulation.py` pass on Windows today and never touch `fmov` with an immediate. Differential testing against Unicorn is what found this, and it is the only thing in the tree that could have.

### DIFF
--- a/src/libtriton/CMakeLists.txt
+++ b/src/libtriton/CMakeLists.txt
@@ -375,6 +375,10 @@ if(PYTHON_BINDINGS)
         install(FILES $<TARGET_FILE_DIR:triton>/triton${PYTHON_SUFFIX} DESTINATION ${PYTHON_SITE_PACKAGES})
     else()
         execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "from sys import version_info; print(f'lib/python{version_info[0]}.{version_info[1]}/site-packages')" OUTPUT_VARIABLE PYTHON_SITE_PACKAGES OUTPUT_STRIP_TRAILING_WHITESPACE)
-        install(FILES $<TARGET_FILE_DIR:triton>/triton${PYTHON_SUFFIX} DESTINATION ${CMAKE_INSTALL_PREFIX}/${PYTHON_SITE_PACKAGES})
+        # PYTHON_SITE_PACKAGES is already relative here, and install() resolves a
+        # relative DESTINATION against CMAKE_INSTALL_PREFIX. Prefixing it again made
+        # the path absolute, which CMake >= 4 rejects under -Werror=dev and which
+        # also defeated `cmake --install --prefix <other>`.
+        install(FILES $<TARGET_FILE_DIR:triton>/triton${PYTHON_SUFFIX} DESTINATION ${PYTHON_SITE_PACKAGES})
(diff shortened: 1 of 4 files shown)


mahmoudimus:
Heads-up on the red checks here: they are **not** caused by this PR.

The failing job is `Build VCPKG package`, and it has failed on every run of that workflow since 2026-06-27 — including [`patch-1`](https://github.com/JonathanSalwan/Triton/actions/runs/31577373781) on 2026-08-12, which predates this branch. The cause is an absolute `install(DESTINATION)` in `src/libtriton/CMakeLists.txt` that only becomes an error on **CMake >= 4**; the workflow uses `lukka/get-cmake@latest`, so the runner moved to 4.4.2 on its own and nothing in the repository changed.

The line number moves between runs only because commits above it shift it: `378` on `patch-1`, `385` here. The statement itself is byte-identical to `dev-v1.0`.

#1457 fixes it, and this branch is now rebased on top of that so it can go green. Note that the workflow runs on these branches are currently held at `action_required` pending maintainer approval.

